PR #21, opened 2022-08-04
Created EnvironmentViewMixin to centralize request environment handling

### LINKED ISSUE #20: Add general handling of Team and Environment headers
Many of the views are going to need to handle the team and environment headers, checking that one or the other is present and then making the appropriate Environment object available in the request.  There should be a general mechanism for doing this that all of the views can leverage rather than having bespoke handling in each view.

scott-taubman:
Closes #20 

This PR adds a mixin to handle the X-Team-Id and X-Environment-Id headers so that the logic can be reused across all of the relevant views.


kvig (approved):
Works for me

Issue #12, opened 2022-07-25
Generate OpenAPI from package yaml and create Function instances


scott-taubman:
The `/publish` endpoint allows uses to submit their package to be built and published.  As part of that, we need to parse the package definition yaml file, create Function instances for each function in the package, and then generate an OpenAPI schema for each function (stored in Function.schema).  That schema will be used for things like validating request input and informing the UI what the function looks like.

More details to come...

scott-taubman:
Closing this specific issue for now and will open follow-ons to continue fleshing out the schema generation.

PR #22, opened 2022-08-05
 #12 - Confirmed schema, added execute endpoint.

kvig:
Added the schema generation to the function creation in the build step.
Added a tasking endpoint to exercise the function parameter validation.

Need to rebase after the team/environment id branch is merged, this is currently hardcoded to a specific environment in my setup.

### DIFF
--- a/functionary/builder/utils.py
+++ b/functionary/builder/utils.py
@@ -200,6 +200,7 @@ def _create_package_from_definition(
 
 
 def _create_functions_from_definition(definitions, package: Package):
+    """Creates the functions in the package from the definition file"""
     for function_def in definitions:
         name = function_def.get("name")
         try:
@@ -215,7 +216,9 @@ def _create_functions_from_definition(definitions, package: Package):
         function_obj.save()
 
 
-def _generate_function_schema(name: str, parameters):
+def _generate_function_schema(name: str, parameters) -> str:
+    """Creates a pydantic model from the parameter definitions and returns the schema
+    as a JSON string"""
     type_map = {"int": int, "str": str}
     params_dict = {}
 
--- a/functionary/builder/utils.py
+++ b/functionary/builder/utils.py
@@ -1,5 +1,4 @@
 import io
-import json
 import logging
 import os
 import shutil
@@ -210,8 +209,8 @@ def _create_functions_from_definition(definitions, package: Package):
 
         function_obj.display_name = function_def.get("display_name")
         function_obj.description = function_def.get("description")
-        function_obj.schema = json.loads(
-            _generate_function_schema(name, function_def.get("parameters"))
+        function_obj.schema = _generate_function_schema(
+            name, function_def.get("parameters")
         )
         function_obj.save()
 
@@ -219,6 +218,7 @@ def _create_functions_from_definition(definitions, package: Package):
 def _generate_function_schema(name: str, parameters) -> str:
     """Creates a pydantic model from the parameter definitions and returns the schema
     as a JSON string"""
+    # TODO - Update this map to add lists and whatever other pieces we want to support
     type_map = {"int": int, "str": str}
     params_dict = {}
 
@@ -234,4 +234,4 @@ def _generate_function_schema(name: str, parameters) -> str:
 
     model = create_model(name, **params_dict)
 
-    return model.schema_json(indent=2)
+    return model.schema()
--- a/functionary/builder/utils.py
+++ b/functionary/builder/utils.py
@@ -1,4 +1,5 @@
 import io
+import json
 import logging
 import os
 import shutil
@@ -8,20 +9,20 @@ from uuid import UUID
 import docker
 import yaml
 from celery.utils.log import get_task_logger
+from django.apps import apps
 from django.conf import settings
 from django.db import transaction
 from django.template import Context, Engine
+from pydantic import Field, create_model
 
-from core.models import Environment, Package, User
+from core.models import Environment, Function, Package, User
 
 from .celery import app
 from .models import Build, BuildResource
 
 _docker_client = docker.from_env()
 
-# TODO: Is there a way to direct access the app dir rather than going through the
-#       project BASE_DIR?
-_dockerfile_home = f"{settings.BASE_DIR}/../builder/resources/docker"
+_dockerfile_home = f"{apps.get_app_config('builder').path}/resources/docker"
 
 logger = get_task_logger(__name__)
 logger.setLevel(getattr(logging, settings.LOG_LEVEL))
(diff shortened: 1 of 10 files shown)


functionary/core/api/v1/views/execute.py
```diff
@@ -0,0 +1,69 @@
...
+    """View for executing a function with specified parameters."""
+
+    serializer_class = ExecuteSerializer
+    # TODO: Proper permissions
+    permission_classes = [permissions.IsAuthenticated]
+
+    @extend_schema(
+        description=("Execute a Function"),
+        request=ExecuteSerializer,
+        responses={200: None},
```
scott-taubman, line 30: This response should probably be 201 (Created).

functionary/core/api/v1/views/execute.py
```diff
@@ -0,0 +1,69 @@
...
+        parameters=EnvironmentViewMixin.header_parameters,
+    )
+    def post(self, request: Request) -> Response:
+        """Accepts a package and function name and executes it with the passed
+        parameters. The parameters are validated against the schema stored with the
+        function prior to execution."""
+        serial = ExecuteSerializer(data=request.data)
```
scott-taubman, line 37: I'm a fan of absurdly long names for things, so let's not truncate in cases like this.  Just call it `serializer`.

functionary/core/api/v1/views/execute.py
```diff
@@ -0,0 +1,69 @@
...
+        serial.is_valid(raise_exception=True)
+
+        package = serial.validated_data["package"]
+        function = serial.validated_data["function"]
+
+        logger.info("Received request to call %s in %s", function, package)
+
+        environment = self.get_environment()
+        try:
+            pack = Package.objects.get(environment_id=environment.id, name=package)
```
scott-taubman, line 47: For consistency with similar assignments elsewhere, let's call this `package_obj`.

functionary/core/api/v1/views/execute.py
```diff
@@ -0,0 +1,69 @@
...
+        except Package.DoesNotExist:
+            raise BadRequest(f"Package not found: {package}")
+
+        try:
+            func = Function.objects.get(package=pack.id, name=function)
```
scott-taubman, line 52: `func` --> `function_obj`

functionary/core/api/v1/views/execute.py
```diff
@@ -0,0 +1,69 @@
...
+            raise BadRequest(f"Function not found: {function}")
+
+        try:
+            jsonschema.validate(
+                instance=serial.validated_data["parameters"], schema=func.schema
+            )
+        except jsonschema.exceptions.ValidationError as ve:
+            raise BadRequest(ve.message)
+
+        logger.info(f"Validation passed, executing {func.name}")
```
scott-taubman, line 63: This feels more like a debug message than info.  And honestly even as a debug I'm not sure it buys us much.  Maybe just drop it and we'll add debug logging as needed over time when we discover the places that they would be beneficial.

functionary/core/api/v1/views/execute.py
```diff
@@ -0,0 +1,69 @@
...
+        """Accepts a package and function name and executes it with the passed
+        parameters. The parameters are validated against the schema stored with the
+        function prior to execution."""
+        serial = ExecuteSerializer(data=request.data)
+        serial.is_valid(raise_exception=True)
+
+        package = serial.validated_data["package"]
+        function = serial.validated_data["function"]
+
+        logger.info("Received request to call %s in %s", function, package)
```
scott-taubman, line 43: This is another one that feels like it's either a debug log message or just goes away for now.

functionary/core/api/v1/views/execute.py
```diff
@@ -0,0 +1,69 @@
...
+        except jsonschema.exceptions.ValidationError as ve:
+            raise BadRequest(ve.message)
+
+        logger.info(f"Validation passed, executing {func.name}")
+
+        stat = status.HTTP_200_OK
+
+        # TODO Fire off tasking request
+
+        return Response(status=stat)
```
scott-taubman, line 69: Not sure you need to store off the status.  You could just make this:

```
return Response(status=status.HTTP_201_CREATED)
```

And as soon as I type that, I feel like I've maybe answered my own question about whether this should be a model viewset for Task.  If we're returning a 201 here, what's the content that the caller might expect to get back?  It's presumably the Task object with the ID of the task so that they can go look up the results later.  Don't worry about changing anything now.  We can move all of this once we have a Task model and all that.  But I have a feeling that's where all of this is going to end up living eventually.

functionary/builder/utils.py
```diff
@@ -192,7 +193,45 @@ def _create_package_from_definition(
...
+        field.title = parameter.get("display_name", field.alias)
+        field.description = parameter.get("description")
+        field.default = parameter.get("default", ...)
+        type_ = type_map[parameter["type"]]
+
+        params_dict[field.alias] = (type_, field)
+
+    model = create_model(name, **params_dict)
+
+    return model.schema_json(indent=2)
```
scott-taubman, line 237: You should just be able to do `model.schema()` here and get it back as a dict.  That'll save you from having to handle the json later.

functionary/builder/utils.py
```diff
@@ -192,7 +193,45 @@ def _create_package_from_definition(
...
+    for function_def in definitions:
+        name = function_def.get("name")
+        try:
+            function_obj = Function.objects.get(package=package, name=name)
+        except Function.DoesNotExist:
+            function_obj = Function(package=package, name=name)
+
+        function_obj.display_name = function_def.get("display_name")
+        function_obj.description = function_def.get("description")
+        function_obj.schema = json.loads(
```
scott-taubman, line 213: As noted below, if you have `_generate_function_schema` return a dict you can save yourself the `json.loads` here.

functionary/builder/utils.py
```diff
@@ -192,7 +193,45 @@ def _create_package_from_definition(
...
+            _generate_function_schema(name, function_def.get("parameters"))
+        )
+        function_obj.save()
+
+
+def _generate_function_schema(name: str, parameters) -> str:
+    """Creates a pydantic model from the parameter definitions and returns the schema
+    as a JSON string"""
+    type_map = {"int": int, "str": str}
```
scott-taubman, line 222: We definitely need at least a TODO to flesh out this type map.

scott-taubman (changes requested):
On the whole looks good.  Most of the comments are about naming and other minor things.

functionary/core/api/v1/views/execute.py
kvig: noted for later

functionary/core/api/v1/views/execute.py
kvig: done

functionary/core/api/v1/views/execute.py
kvig: done

functionary/core/api/v1/views/execute.py
kvig: done

functionary/core/api/v1/views/execute.py
kvig: gone

functionary/core/api/v1/views/execute.py
kvig: Changed to debug, I'd at least like to know I"m hitting the endpoint with something potentially valid

functionary/core/api/v1/views/execute.py
kvig: Noted for later

functionary/builder/utils.py
kvig: done

functionary/builder/utils.py
kvig: done

functionary/builder/utils.py
kvig: done

scott-taubman (approved):

PR #24, opened 2022-08-09
#23 - Roles and Permissions

### LINKED ISSUE #23: Add environment memberships
We want users to be able to have roles in specific Environments, not just at the team level.  Users would then be considered to have the superset of roles that they are assigned at the team and Environment levels.  That is, a user assigned roles for an Environment consist of the roles they are explicitly assigned for that Environment, plus the roles that they are assigned for the Team to which the Environment belongs.

scott-taubman:
Closes #23 

This PR does the following:

* Replaces the `TeamMember` model with two separate models: `TeamUserRole` and `EnvironmentUserRole`.
* Establishes an initial set of Roles and their corresponding permissions, which are just based on CRUD operations for each of the models for now.
* Implements a custom auth backend to allow checking if a user has a given permission for a Team or Environment, based on their assigned `TeamUserRoles` and `EnvironmentUserRoles`.

Note that this PR does not yet implement permissions checks into any of the views.  I avoided doing this for now as to not make this PR grow excessively large.

## Testing Instructions
To see this in action, first user the django admin pages to do the following:
* Create a non-superuser account
* Create a Team and any number of environments.
* Create TeamUserRole and EnvironmentUserRole entries for the user.

Now launch a django shell and do something like the following:
```python
from core.auth import Permissions
from core.models import User, Environment, Team

# In this example, snake has the READ_ONLY role for the foxhound default environment
snake = User.objects.get(username="snake")
team = Team.objects.get(name="foxhound")
environment = Environment.objects.get(team=team, name="default")

snake.team_permissions(team)
# set()

snake.environment_permissions(environment)
"""
{'environment:read',
 'function:read',
 'package:read',
 'team:read',
 'user:read',
 'userrole:read'}
"""

snake.has_perm(Permissions.PACKAGE_READ, environment)
# True

snake.has_perm(Permissions.PACKAGE_READ, team)
# False

snake.has_perm(Permissions.PACKAGE_CREATE, environment)
# False
```

If we granted snake the ADMIN role at the foxhound team level, we would see this:

```python
snake.team_permissions(team)
"""
{'environment:create',
 'environment:delete',
 'environment:read',
 'environment:update',
 'function:create',
 'function:delete',
 'function:read',
 'function:update',
 'package:create',
 'package:delete',
 'package:read',
 'package:update',
 'team:create',
 'team:delete',
 'team:read',
 'team:update',
 'user:create',
 'user:delete',
 'user:read',
 'user:update',
 'userrole:create',
 'userrole:delete',
 'userrole:read',
 'userrole:update'}
"""

snake.environment_permissions(environment)
"""
{'environment:read',
 'function:read',
 'package:read',
 'team:read',
 'user:read',
 'userrole:read'}
"""

snake.environment_permissions(environment, inherited=True)
"""
{'environment:create',
 'environment:delete',
 'environment:read',
 'environment:update',
 'function:create',
 'function:delete',
 'function:read',
 'function:update',
 'package:create',
 'package:delete',
 'package:read',
 'package:update',
 'team:create',
 'team:delete',
 'team:read',
 'team:update',
 'user:create',
 'user:delete',
 'user:read',
 'user:update',
 'userrole:create',
 'userrole:delete',
 'userrole:read',
 'userrole:update'}
"""

snake.has_perm(Permissions.PACKAGE_CREATE, team)
# True

snake.has_perm(Permissions.PACKAGE_CREATE, environment)
# True
```


functionary/README.md
```diff
@@ -1,11 +1,11 @@
 # Functionary
 
-## Setup:
+## Setup
```
scott-taubman, line 3: Unrelated changes in this file, just minor updates that should be here.

functionary/builder/models/build.py
```diff
@@ -32,7 +32,7 @@ class Build(models.Model):
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     created_at = models.DateTimeField(auto_now_add=True, db_index=True)
-    updated_at = models.DateTimeField(auto_now_add=True, db_index=True)
+    updated_at = models.DateTimeField(auto_now=True, db_index=True)
```
scott-taubman, line 35: Unrelated, but I noticed this was wrong.

functionary/core/apps.py
```diff
@@ -1,6 +1,6 @@
 from django.apps import AppConfig
 
 
-class TaskmanConfig(AppConfig):
-    default_auto_field = 'django.db.models.BigAutoField'
-    name = 'core'
+class CoreConfig(AppConfig):
```
scott-taubman, line 4: Another unrelated item.  Left over reference to the name in the proof of concept phase.

functionary/core/auth/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+
+    ENVIRONMENT_CREATE = "environment:create"
+    ENVIRONMENT_READ = "environment:read"
+    ENVIRONMENT_UPDATE = "environment:update"
+    ENVIRONMENT_DELETE = "environment:delete"
+
+    FUNCTION_CREATE = "function:create"
+    FUNCTION_READ = "function:read"
+    FUNCTION_UPDATE = "function:update"
+    FUNCTION_DELETE = "function:delete"
```
kvig, line 29: What are the constraints for a user to execute a function? If they can see it and have an environment, they can execute it?

functionary/core/auth/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+
+# READ_ONLY gets read access to all models
+_READ_ONLY_PERMISSIONS = [
+    permission.value for permission in Permissions if ":read" in permission.value
+]
+
+# Other roles get build on READ_ONLY
+_DEVELOPER_PERMISSIONS = _READ_ONLY_PERMISSIONS + [
+    Permissions.PACKAGE_CREATE.value,
+    Permissions.PACKAGE_UPDATE.value,
```
kvig, line 72: FUNCTION_CREATE and FUNCTION_UPDATE too?

kvig (commented):
Couple comments. Roles displayed and then updated as described when testing.

functionary/core/auth/__init__.py
scott-taubman: Right now I'm still thinking that at some point soon we'll have a `Task` model, which would have its own set of CRUD permissions here.  So executing a function would be governed by the `task:create` permission.

functionary/core/auth/__init__.py
scott-taubman: I created the CRUD permissions for each of the models, but I'm not sure yet that we'll use them all.  With the way the package publishing process works, it feels like it makes sense to just have `PACKAGE_CREATE` decide if you can `POST` to the `/api/v1/publish` endpoint.  And then obviously once you do that, functions are generated as part of that process.  For now I don't see an advantage to checking for `PACKAGE_CREATE` and `FUNCTION_CREATE` independently, but if that changes as we're implementing things and thinking about different scenarios, it's easy enough to modify the list of permissions for any given role.

kvig (approved):
Ship it!

PR #26, opened 2022-08-09
#25 - Add Environment to Build model and improve publish exception handling

### LINKED ISSUE #25: Add Environment to Build model
Currently the environment id is being passed into the build celery task.  What we needs instead is for the Build model to have the Environment stored on it to facilitate permissions checks.

scott-taubman:
Closes #25 

The PR does the following:
* Adds Environment as a foreign key on the Build model.
* Updates the build task code to leverage that Environment info on the Build, rather than requiring an environment_id be passed in.
* Streamlines the template loading for the Dockerfile.
* Improves the exception handling on the `/api/v1/publish` endpoint so that 400 responses with useful messages are returned in more situations.

## Testing Instructions

Try publishing a package to the `/api/v1/publish` endpoint.

For a valid package tarball, everything should work the same.  The only difference will be that there is an environment field on the resulting build object.

For various incarnations of invalid package tarballs, you should now get a 400 response with a semi-useful message from the `/api/v1/publish` endpoint rather than a 500 error.  For example:

* Providing a file that is not a tar file at all
* Providing a tar file that doesn't have a package.yaml
* The package.yaml is present, but improperly formatted

All of these would have resulted in either a 500 error before, or maybe a 400 but with less context in the resulting message than is present now.


kvig (approved):
I like the template engine change and selecting the environment with the build.

Tried out the publish, also accidentally exercised an error path. Looks good.

PR #27, opened 2022-08-10
#17 - Update CLI login and publish commands

### LINKED ISSUE #17: Update CLI login and publish
The CLI needs to be updated to be in line with recent work on main app.

* The login command needs to hit the /api/v1/api-token-auth endpoint to retrieve the token.
* The header that gets set to use the token needs to prefix the token with "Token" rather than "Bearer"
* The package tarball needs to put the files at the root of the structure rather than nested in a subdirectory
* The package tarball needs to be sent to /api/v1/publish and the request must include the `X-Environment-Id` header.

catattack05:
Closes #17 

-The login command now hits the /api/v1/api-token-auth endpoint to retrieve the token.
-The header that gets set to use the token prefixes the token with "Token" rather than "Bearer"
-The package tarball puts the files at the root of the structure rather than nested in a subdirectory
--The package tarball is sent to /api/v1/publish and the request includes the X-Environment-Id header.

### DIFF
--- a/cli/functionary/package.py
+++ b/cli/functionary/package.py
@@ -5,6 +5,7 @@ import tarfile
 import click
 import requests
 import yaml
+import os
 
 from .tokens import TokenError, getToken
 
@@ -24,6 +25,11 @@ def generateYaml(output_dir: str, name: str, language: str):
     path = pathlib.Path(output_dir).resolve() / name / f"{name}.yaml"
     with path.open(mode="w"):
         path.write_text(yaml.dump(metadata))
+
+
+def get_environment_id():
+    environment_id = os.environ.get("FUNCTIONARY_ENVIRONMENT")
+    return environment_id
 
 
 @click.group("package")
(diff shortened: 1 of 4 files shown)


cli/functionary/package.py
```diff
@@ -26,6 +29,10 @@ def generateYaml(output_dir: str, name: str, language: str):
         path.write_text(yaml.dump(metadata))
 
 
+def getEnvironmentID():
```
scott-taubman, line 32: This should be `get_environment_id`.  I realize that the function immediately before this fails to follow the correct [function naming convention](https://peps.python.org/pep-0008/#function-and-variable-names) as well, which is on me since I reviewed that code too.  But let's fix this one (and `environmentId` below) and work toward some consistency.

cli/functionary/package.py
```diff
@@ -82,20 +89,22 @@ def publish(ctx, path, host):
...
     # publish should http the tar to a server, wait for return
     upload_file = open(tarfile_name, "rb")
     upload_response = None
-    headers = {"Authorization": f"Bearer {token}"}
-
+    environment_id = str(getEnvironmentID())
+    print('Token ' + token)
+    print('ID ' + environment_id)
+    headers = {"Authorization": f"Token {token}", "X-Environment-ID": f"{environment_id}"}
+    publish_host = host + "/api/v1/publish"
```
scott-taubman, line 104: I'm really picking nits here, but this should probably be `publish_url` rather than `publish_host` since it consists of more than just the host info.

scott-taubman (changes requested):
Works as intended.  Comments are mostly related to formatting, naming conventions, and misc cleanup.

cli/functionary/tokens.py
```diff
@@ -50,12 +47,11 @@ def saveTokens(tokens):
     with tokensFile.open("wt"):
         tokensFile.write_text(
             f"token={tokens['token']}\n"
-            f"login_url={tokens['login_url']}"
         )
 
 
 def getToken():
-    tokensFile = Path.home() / ".functionary" / "tokens"
+    tokensFile = Path.home() / ".functionary" / "tokens"
```
scott-taubman, line 54: Still have some white space here.

cli/functionary/package.py
```diff
@@ -26,6 +28,10 @@ def generateYaml(output_dir: str, name: str, language: str):
         path.write_text(yaml.dump(metadata))
 
 
+def get_environment_id():
+    environmentId = os.environ.get('FUNCTIONARY_ENVIRONMENT')
```
scott-taubman, line 32: You changed the function name, but this needed to changed to.  Or you can just return the `os.environ.get` call directly and drop the variable.

scott-taubman (changes requested):

scott-taubman (approved):

PR #33, opened 2022-08-15
#29 - Store host after login

### LINKED ISSUE #29: CLI - Store host after login
The functionary host should only need to be provided when running the login command and not on every subsequent command.

## How it currently works
Right now, logging in only stores the token.  So we do something like:
```shell
$ functionary login http://localhost:8000
```
And then when you want to publish a package, you have to supply the host again:
```shell
$ functionary package publish /path/to/mypackage http://localhost:8000
```
This is rather redundant and annoying

## How it could work
Instead, of the above we should store the host that we're logged into, not just the token.  Then that host info would be used on all subsequent commands.

```shell
$ functionary login http://localhost:8000

# Now I don't need to supply the host on the publish command
$ functionary package publish /path/to/mypackage
```

## Implementation Notes
* Doing this requires that we expand the stored config for the CLI. Currently we write down a token file in ~/.functionary/tokens. We should change the format (and name) of this so that it's a generic config file instead. I mentioned this in #28 as well, so have that issue in mind as well when thinking about the new format for the config.
* For now, assume that a user is only ever logged into one instance of functionary at a time.  We don't need to complicate things at this point by supporting multiple hosts.  If the user is logged into one host and then runs the login command again for a different host, it's acceptable at this point if that results in the config for the previous host being overwritten.

catattack05:
Closes #29 

I've duplicated a lot of what we did to store the tokens, though both tokens and the host are now stored in the same config file. I also am trying to stick with camel case for things.

### DIFF
--- a/cli/functionary/config.py
+++ b/cli/functionary/config.py
@@ -37,5 +37,5 @@ def get_config_value(parameter):
         }
         return config[parameter]
     # if path not found or key not found, raise error
-    except KeyError as e:
+    except KeyError:
         raise click.ClickException("Could not find config parameter")
--- a/cli/functionary/config.py
+++ b/cli/functionary/config.py
@@ -1,0 +1,41 @@
+from pathlib import Path
+
+import click
+from dotenv import dotenv_values
+
+
+def save_config_value(parameter, value):
+    functionary_dir = Path.home() / ".functionary"
+    if not functionary_dir.exists():
+        functionary_dir.mkdir()
+
+    config_file = functionary_dir / "config"
+
+    # if the parameter does not exist already, write it
+    if parameter not in dotenv_values(config_file):
+        with config_file.open("a") as f:
+            f.write(f"{parameter}={value}\n")
+
+    else:  # if the parameter already has a value, overwrite it
+        with open(config_file, "r", encoding="utf-8") as file:
+            data = file.readlines()
+            index = -1
+            for line in data:
+                if parameter in line:
+                    index = data.index(line)
+            data[index] = f"{parameter}={value}\n"
+        with open(config_file, "w", encoding="utf-8") as file:
+            file.writelines(data)
+
+
+def get_config_value(parameter):
+    try:
+        config_file = Path.home() / ".functionary" / "config"
+
+        config = {
+            **dotenv_values(str(config_file)),
+        }
+        return config[parameter]
+    # if path not found or key not found, raise error
+    except KeyError as e:
+        raise click.ClickException("Could not find config parameter")
--- a/cli/functionary/config.py
+++ b/cli/functionary/config.py
@@ -10,13 +10,14 @@ def save_config_value(key, value):
     overwrite the current value.
 
     Args:
-        key and value to be stored as strings
+        key: the configuration parameter to set
+        value: the value to set the configutation parameter to
 
     Returns:
-        Nothing
+        None
 
     Raises:
-        ClickException from PermissionError if cannot read config file
+        ClickException: Received a PermissionError when opening config file
     """
     try:
         functionary_dir = Path.home() / ".functionary"
@@ -30,7 +31,7 @@ def save_config_value(key, value):
             value,
         )
     except PermissionError:
-        raise click.ClickException("Config file present, but could not be read")
+        raise click.ClickException(f"Failed to open {config_file}: Permission Denied")
 
 
 def get_config_value(key):
@@ -38,14 +39,14 @@ def get_config_value(key):
     Retrieve the value associated with a key from the config file
 
     Args:
-        The key to be found
+        key: the configuration parameter to retrieve the value of
 
     Returns:
         Value associated with that key as a string
 
     Raises:
-        ClickException if value returned is None (key doesn't exist in file)
-        ClickException from PermissionError if file cannot be read
+        ClickException: Received PermissionError when opening file or no configuration
+        parameter matching the provided key exists
     """
     try:
         config_file = Path.home() / ".functionary" / "config"
@@ -56,4 +57,4 @@ def get_config_value(key):
             return value
     # if path not found or key not found, raise error
     except PermissionError:
-        raise click.ClickException("Config file present, but could not be read")
+        raise click.ClickException(f"Failed to open {config_file}: Permission Denied")
--- a/cli/functionary/config.py
+++ b/cli/functionary/config.py
@@ -5,26 +5,55 @@ from dotenv import get_key, set_key
 
 
 def save_config_value(key, value):
-    functionary_dir = Path.home() / ".functionary"
-    if not functionary_dir.exists():
-        functionary_dir.mkdir()
+    """
+    Save configuration key and value to config file. If key is already present,
+    overwrite the current value.
 
-    config_file = functionary_dir / "config"
-    set_key(
-        config_file,
-        key,
-        value,
-        quote_mode="always",
-        export=False,
-        encoding="utf-8",
-    )
+    Args:
+        key and value to be stored as strings
+
+    Returns:
+        Nothing
+
+    Raises:
+        ClickException from PermissionError if cannot read config file
+    """
+    try:
+        functionary_dir = Path.home() / ".functionary"
+        if not functionary_dir.exists():
+            functionary_dir.mkdir()
+
+        config_file = functionary_dir / "config"
+        set_key(
+            config_file,
+            key,
+            value,
+        )
+    except PermissionError:
+        raise click.ClickException("Config file present, but could not be read")
 
 
 def get_config_value(key):
+    """
+    Retrieve the value associated with a key from the config file
+
+    Args:
+        The key to be found
+
+    Returns:
+        Value associated with that key as a string
+
+    Raises:
+        ClickException if value returned is None (key doesn't exist in file)
+        ClickException from PermissionError if file cannot be read
+    """
     try:
         config_file = Path.home() / ".functionary" / "config"
         value = get_key(config_file, key)
-        return value
+        if value is None:
+            raise click.ClickException(f"Could not find value for {key}")
+        else:
+            return value
     # if path not found or key not found, raise error
-    except KeyError:
-        raise click.ClickException("Could not find config parameter")
+    except PermissionError:
+        raise click.ClickException("Config file present, but could not be read")
--- a/cli/functionary/config.py
+++ b/cli/functionary/config.py
@@ -1,41 +1,30 @@
 from pathlib import Path
 
 import click
-from dotenv import dotenv_values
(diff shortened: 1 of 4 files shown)


cli/functionary/login.py
```diff
@@ -1,8 +1,15 @@
+from pathlib import Path
+
 import click
+from dotenv import dotenv_values
 
 from .tokens import login
 
 
+class host_error(Exception):
```
scott-taubman, line 9: This exception (and the token one) aren't needed.

cli/functionary/login.py
```diff
@@ -21,10 +28,33 @@ def login_cmd(ctx, user, password, host):
...
     else:
         click.secho(
             message,
             err=True,
             fg="red",
         )
         ctx.exit(1)
+
+
+def save_host(host):
```
scott-taubman, line 43: This duplicates the similar call for the token.  Instead, we should refactor this so that there is a generic set of config handling functions.  Probably in a `config.py`, such that you can just do something like:

```
get_config("host")
set_config("host", "localhost:8000")
```

scott-taubman: Also, you'll want exception handling for situations such as the config file not being able to be read or written to.  In such cases, raise a `ClickException` because that will handle the output and exit code for you and save you from having to handle these exceptions in whatever code is calling your config handling functions.

cli/functionary/config.py
```diff
@@ -0,0 +1,41 @@
+from pathlib import Path
+
+import click
+from dotenv import dotenv_values
+
+
+def save_config_value(parameter, value):
```
scott-taubman, line 7: Since we're using dotenv (at least for now), take a look at the [documentation](https://saurabh-kumar.com/python-dotenv/reference/).  I suspect that `set_key` is the thing you want here.

cli/functionary/config.py
```diff
@@ -0,0 +1,41 @@
...
+        with open(config_file, "w", encoding="utf-8") as file:
+            file.writelines(data)
+
+
+def get_config_value(parameter):
+    try:
+        config_file = Path.home() / ".functionary" / "config"
+
+        config = {
+            **dotenv_values(str(config_file)),
```
scott-taubman, line 36: You probably just want to use `get_key` here (see my documentation comment), but I'll mention a couple of things anyway:

* `dotenv_values` can take a "Pathlike" object, so you don't need to convert it to a string first.
* It also returns a dict.

Basically this line could just be:

```python
config = dotenv_values(config_file)
```

Just pointing that out for educational purposes.  Look at using `get_key` instead.

cli/functionary/login.py
```diff
@@ -16,15 +17,7 @@ def login_cmd(ctx, user, password, host):
...
-    if success:
-        click.echo("Login successful!")
-    else:
-        click.secho(
-            message,
-            err=True,
-            fg="red",
-        )
-        ctx.exit(1)
+    # try to login, click will raise error from inside login if something goes wrong
```
scott-taubman, line 20: This comment isn't really needed.

cli/functionary/login.py
```diff
@@ -16,15 +17,7 @@ def login_cmd(ctx, user, password, host):
...
+    login(login_url, user, password)
+    click.echo("Login successful!")
+    save_config_value("host", host)
```
scott-taubman, line 23: There's a bit of inconsistency here in that the token value gets saved inside the `login` call, but the host is saved here after the call.  It seems like the desired behavior is the same: If the login is successful, save the config values.

I think what I would probably do is make it so that the login function doesn't have to concern itself with the config.  Instead, make it return the token upon successful login.  Then save both the token and the host here.

cli/functionary/package.py
```diff
@@ -107,20 +102,17 @@ def publish(ctx, path, host):
             publish_url, headers=headers, files={"package_contents": upload_file}
         )
     except requests.ConnectionError:
-        click.echo(f"Unable to connect to {host}")
-        ctx.exit(2)
+        click.ClickException(f"Unable to connect to {host}")
```
scott-taubman, line 105: The `raise` is missing here and in a few other spots.

cli/functionary/package.py
```diff
@@ -107,20 +102,17 @@ def publish(ctx, path, host):
...
     else:
-        click.echo(
+        click.ClickException(
             f"Failed to build image: {upload_response.status_code}\n"
             f"\tResponse: {upload_response.text}"
         )
         if upload_response.status_code == 401:
-            click.echo("\n\nTry log in again.")
-        ctx.exit(1)
+            click.ClickException("\n\nTry log in again.")
```
scott-taubman, line 118: May as well fix the bad English here.  Also, I'm not sure that the flow here is going to actually work.  Presumably the `click.ClickException` above is support to be raised, which means you'd never reach the `if upload_response.status_code == 401` block.  Take another look at what this whole `else` block is trying to achieve and see if you can correctly rework it while using the `ClickException`.

scott-taubman (changes requested):

cli/functionary/config.py
```diff
@@ -0,0 +1,30 @@
...
+        functionary_dir.mkdir()
+
+    config_file = functionary_dir / "config"
+    set_key(
+        config_file,
+        key,
+        value,
+        quote_mode="always",
+        export=False,
+        encoding="utf-8",
```
scott-taubman, line 19: It doesn't hurt anything the way you have it, but when you see this function documented as:

```python
set_key(
    dotenv_path,
    key_to_set,
    value_to_set,
    quote_mode="always",
    export=False,
    encoding="utf-8",
)
```

Those last 3 have default values set, meaning they are optional parameters.  Since you're just using the default values, you could have left them off and just made this `set_key(config_file, key, value)`.

cli/functionary/config.py
```diff
@@ -0,0 +1,30 @@
+from pathlib import Path
+
+import click
+from dotenv import get_key, set_key
+
+
+def save_config_value(key, value):
```
scott-taubman, line 7: You'll still want to add docstrings and type hinting to these functions.

cli/functionary/config.py
```diff
@@ -0,0 +1,30 @@
...
+    )
+
+
+def get_config_value(key):
+    try:
+        config_file = Path.home() / ".functionary" / "config"
+        value = get_key(config_file, key)
+        return value
+    # if path not found or key not found, raise error
+    except KeyError:
```
scott-taubman, line 29: With the changes you made, I don't believe KeyError will ever be thrown here.

`get_key` returns `None` if the provided file doesn't exist or the key is not present in the file.  You'll want to handle that, otherwise you'll be returning `None` and something will blow up downstream because it was expecting a string to come back.

Also, if the config file is present, but can't be read, it looks like a `PermissionError` gets thrown, so you'll want to handle that (both here and in the set function) so that you can return a nice error rather than just spewing a stack trace to the screen.

cli/functionary/tokens.py
```diff
@@ -1,58 +1,25 @@
...
-            saveTokens(tokens)
-            success = True
+            token = json.loads(login_response.text).get("token")
+            return token
         else:
-            message = (
+            raise click.ClickException(
                 f"Failed to login: {login_response.status_code}\n"
-                f"\tResponse: {login_response.text}"
+                + f"Response: {login_response.text}"
```
scott-taubman, line 20: Super minor thing, but you don't need the `+` here.  The strings will get concatenated.

cli/functionary/login.py
```diff
@@ -16,15 +17,7 @@ def login_cmd(ctx, user, password, host):
...
-        click.echo("Login successful!")
-    else:
-        click.secho(
-            message,
-            err=True,
-            fg="red",
-        )
-        ctx.exit(1)
+    token = login(login_url, user, password)
+    click.echo("Login successful!")
```
scott-taubman, line 21: This message should come after the `save_config_value` calls, since if one of those calls fails, the fact that the login succeeded doesn't really help.

scott-taubman (changes requested):

cli/functionary/config.py
```diff
@@ -0,0 +1,59 @@
...
+import click
+from dotenv import get_key, set_key
+
+
+def save_config_value(key, value):
+    """
+    Save configuration key and value to config file. If key is already present,
+    overwrite the current value.
+
+    Args:
```
scott-taubman, line 12: For consistency with the style we use, the two arguments should be listed separately like:

```
Args:
    key: the configuration parameter to set
    value: the value to set the configuration parameter to
```

cli/functionary/config.py
```diff
@@ -0,0 +1,59 @@
...
+        key and value to be stored as strings
+
+    Returns:
+        Nothing
```
scott-taubman, line 16: None

cli/functionary/config.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Raises:
+        ClickException from PermissionError if cannot read config file
```
scott-taubman, line 19: Again for consistency, this should be:

```
Raises:
    ClickException:  Received a PermissionError when opening the config file
```

cli/functionary/config.py
```diff
@@ -0,0 +1,59 @@
...
+    except PermissionError:
+        raise click.ClickException("Config file present, but could not be read")
+
+
+def get_config_value(key):
+    """
+    Retrieve the value associated with a key from the config file
+
+    Args:
+        The key to be found
```
scott-taubman, line 41: Same comment as earlier.  Use the format:

```
key: The configuration parameter to retrieve the value of
```

cli/functionary/config.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Returns:
+        Value associated with that key as a string
+
+    Raises:
+        ClickException if value returned is None (key doesn't exist in file)
```
scott-taubman, line 47: You can just list ClickException once here:

```
Raises:
    ClickException: A PermissionError was encountered or no configuration parameter matching the provided key exists
```

cli/functionary/config.py
```diff
@@ -0,0 +1,59 @@
...
+            functionary_dir.mkdir()
+
+        config_file = functionary_dir / "config"
+        set_key(
+            config_file,
+            key,
+            value,
+        )
+    except PermissionError:
+        raise click.ClickException("Config file present, but could not be read")
```
scott-taubman, line 33: In this case, the error is writing the file rather than reading.  But in either case, I think both this and the other instance of this message could be:

```
raise click.ClickException(f"Failed to open {config_file}: Permission Denied")
```

That more clearly communicates that the user should go check the permissions on a specific file.

scott-taubman (changes requested):

scott-taubman (approved):

PR #36, opened 2022-08-16
#34 - Environment ViewSets

### LINKED ISSUE #34: Environment based viewsets
Many of the endpoints are intended to always be accessed in the context of a specific environment.  We should create a ViewSet class that handles all of the queryset filtering, header parameter documentation, etc.  We would then update any ViewSets building off of the *ModelViewSet classes to build off of our custom one instead such that:

* Data returned is automatically filtered down to that relevant for the provided environment or team id.
* Documentation of the environment and team headers is in one place, rather than having to be documented on every viewset.

scott-taubman:
Closes #34 

This PR adds `EnvironmentModelViewSet` and `EnvironmentReadOnlyModelViewSet`, which are intended to be used in place of the DRF `ModelViewSet` and `ReadOnlyModelViewSet` in situations where the ViewSet's queryset should be filtered to objects belonging to the request's environment context.  The following functionality is provided:

* The queryset is automatically filtered to the correct objects for the environment.
* If the relationship to an environment is indirect (i.e. The `Function` model has an environment through it's associated package), that can be defined via the `environment_through_field` attribute.
* The environment and team header parameters are included so that they do not have to be explicitly added to ViewSets building off of these new ones.

## Test Instructions
* Create packages, functions, and builds for two separate environments.
* Query the respective endpoints.  In each case you should only see results that are for the environment ID (or default environment for the team id) that you supplied via the header parameter.  Before these changes you would always see all results, regardless of the provided environment.

### DIFF
--- a/functionary/core/api/mixins.py
+++ b/functionary/core/api/mixins.py
@@ -1,3 +1,5 @@
+from functools import cache
+
 from django.core.exceptions import ValidationError
 from drf_spectacular.utils import OpenApiParameter
 
@@ -29,6 +31,7 @@ class EnvironmentViewMixin:
         ),
     ]
 
+    @cache
     def get_environment(self) -> Environment:
         """Retrieve the Environment object that corresponds to either the environment
         with id X-Environment-Id or the default environment for the team with id
--- a/functionary/README.md
+++ b/functionary/README.md
@@ -53,6 +53,15 @@ location using the environment variables described earlier.
 started working on Functionary. Instructions for general users will be added as
 the project matures.
 
+### Python Requirements
+
+Functionary is being developed with containers being the primary target for
+deployment. As such, the intention is to officially support only the latest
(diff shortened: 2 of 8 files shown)


functionary/core/api/mixins.py
```diff
@@ -29,6 +31,7 @@ class EnvironmentViewMixin:
         ),
     ]
 
+    @cache
```
kvig, line 34: Since this is 3.9+ should we explicitly document our minimum Python version somewhere?

kvig (commented):
Looks good, just one question

functionary/core/api/mixins.py
scott-taubman: I added a small section in the README explaining that the goal is to always be on the latest version of python.

kvig (approved):
LGTM

PR #38, opened 2022-08-16
#37 - Environment based permission class for views

### LINKED ISSUE #37: Environment based endpoint permission
For model based ViewSets we want automate endpoint permission access based on the following request info:

* Action (CRUD)
* Environment
* Model

Using this combination of info, we want a permissions check to happen for the requesting user and deny access when appropriate.

This most like should take the form of a custom DRF permission class.

scott-taubman:
Closes #37 

This PR does the following:

* Sets the drf `DEFAULT_PERMISSION_CLASSES` to `isAuthenticated`.  Any views that intend to use that as their permission class no longer have it explicitly defined.
* Adds a `HasEnvironmentPermissionForAction` permission class for use on viewsets that inherit build off of the `EnvironmentViewMixin` mixin in some way.  This class determines the permission needed for the request and then grants or denies access based on the user's permissions.
* The `ExecuteView` remains untouched for now.  Although it falls into the same category as `PublishView`, that is going to change soon as we move that logic over to a `TaskViewSet`.  The permissions enforcement will be added then.

## Testing Instructions

* Create a non-admin user.  Do not assign it any user roles.
* Hitting the functions, builds, or packages endpoints as that user should result in a 403.
* Assign the user the read only role for some environment.
* Now doing a GET against endpoints should work, but trying to POST (such as the /publish endpoint) should still 403.
* Feel free to play around with other things.

### DIFF
--- a/functionary/builder/api/v1/views/build.py
+++ b/functionary/builder/api/v1/views/build.py
@@ -1,14 +1,14 @@
-from rest_framework import permissions
-
 from builder.models import Build
+from core.api.permissions import HasEnvironmentPermissionForAction
 from core.api.viewsets import EnvironmentReadOnlyModelViewSet
 
 from ..serializers import BuildSerializer
 
 
 class BuildViewSet(EnvironmentReadOnlyModelViewSet):
+    """View the status of package builds"""
+
     queryset = Build.objects.all()
     serializer_class = BuildSerializer
-
-    # TODO: Proper permissions
-    permission_classes = [permissions.IsAuthenticated]
+    permission_classes = [HasEnvironmentPermissionForAction]
+    permissioned_model = "Package"
--- a/functionary/builder/api/v1/views/publish.py
+++ b/functionary/builder/api/v1/views/publish.py
@@ -1,5 +1,4 @@
 from drf_spectacular.utils import extend_schema
-from rest_framework import permissions
 from rest_framework.exceptions import ParseError
 from rest_framework.parsers import MultiPartParser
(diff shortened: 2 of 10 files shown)


functionary/builder/api/v1/views/build.py
```diff
@@ -1,14 +1,14 @@
...
-
 from builder.models import Build
+from core.api.permissions import HasEnvironmentPermissionForAction
 from core.api.viewsets import EnvironmentReadOnlyModelViewSet
 
 from ..serializers import BuildSerializer
 
 
 class BuildViewSet(EnvironmentReadOnlyModelViewSet):
+    """View the status of package builds"""
```
scott-taubman, line 9: The description in the swagger docs was defaulting to that of `EnvironmentReadOnlyModelViewSet`, so I added a very basic docstring for now so that at least the generated swagger docs have something more sensible.  We'll surely flesh out the docs down the line.

scott-taubman:
> Works as described. I was able to publish to one environment that I had developer access to and was prevented from publishing to the default readonly environment for the team.
> 
> My only observation, which is probably for the future, is that before I assigned the user to the team, I was able to see all the environments for the teams. I'm not sure if we want to restrict that, I'm on board with seeing which teams are available and allowing a request to join team, but wasn't sure if the user should see all environments in teams they aren't members of.

The Team endpoint and the User endpoint I was planning on tackling separately, since the implementation of the data filtering on those is going to be different than what's been added here.

PR #40, opened 2022-08-18
#28: CLI - Add env set and list

### LINKED ISSUE #28: CLI - List and set environment
It should be possible for a user to see a list of the available environments and then set one to be the active environment that subsequent commands operate on.

## How it currently works
When publishing a package, for example, you currently need to go lookup the environment id outside of the CLI, and then provide that as an environment variable when running publish.
```shell
$ export FUNCTIONARY_ENVIRONMENT="2cb08859-edca-44b9-9b59-6d274653e211"
$ functionary package publish /path/to/mypackage
```

## How it could work
What we want instead is for the CLI to have a mechanism for listing environments as well as setting an active environment.  It could go something like this:

```shell
$ functionary environment set
Available Environments:
   1) foxhound - dev
   2) foxhound - prod
   3) patriots - default

Select Environment:  2   # this is a prompt where the user enters a number from the list above
Active environment is now "foxhound - prod".
```
In the above, the environments are listed as "team_name - environment_name".  

And the user could list environments, showing the active one, like this:

```shell
$ functionary environment list
  foxhound - dev
* foxhound - prod
  patriots - default
```
Where the "*" indicates the currently selected environment.

Once the environment is set like this, that info would be leveraged but subsequent commands, such as publish:

```shell
# No longer need to set the environment variable
$ functionary package publish /path/to/mypackage
```

## Implementation Notes
* The information displayed in the `functionary environment list` and `functionary environment set` commands would come from the `/api/v1/teams` endpoint.
* Doing this requires that we expand the stored config for the CLI.  Currently we write down a token file in `~/.functionary/tokens`.  We should change the format (and name) of this so that it's a generic config file instead.  The token would be one bit of info that it holds.  The selected environment would be another.  Note that the environment ID needs to be stored, not necessarily the name, as that's what needs to be used by commands such as "publish".
* `functionary environment` is a lot to type.  It would be nice if `env` worked as an alias too, so you could optionally do `functionary env list` instead.

catattack05:
Closes #28 

Created new environment.py file that has a get and a set command, storing the current environment in the config file. Changed publish to use the new config file value instead of us exporting through the command line.

Original issue had wondered about creating an alias so that environment and env would go to the same list of commands. I did a little searching and it doesn't look like Click supports this out of the box, but there do seem to be a few libraries that support it that we could import (like this one for example: [https://pypi.org/project/click-aliases/](url)

### DIFF
--- a/cli/functionary/environment.py
+++ b/cli/functionary/environment.py
@@ -1,0 +1,98 @@
+import json
+
+import click
+import requests
+
+from .config import get_config_value, save_config_value
+
+
+def get_environment_list():
+    """
+    Helper function to get the environment list from host
+
+    Args:
+        None
+
+    Returns:
+        None
+
+    Raises:
+        ClickException if bad get request
+    """
+    token = get_config_value("token")
+    teams_url = get_config_value("host") + "/api/v1/teams"
+    header = {"Authorization": f"Token {token}"}
+    response = requests.get(teams_url, headers=header)
+
+    if response.ok:
+        data = json.loads(response.text).get("results")
+
+        env_list = []
+        for team in data:
+            for env_set in team.get("environments"):
+                env_list.append(env_set)
+        return env_list
+    else:
+        raise click.ClickException(
+            f"Failed to get environment list: {response.status_code}\n"
+            f"Response: {response.text}"
+        )
+
+
+@click.group("environment")
+@click.pass_context
+def environment_cmd(ctx):
+    pass
+
+
+@environment_cmd.command()
+@click.pass_context
+def set(ctx):
+    """
+    Command to set the environment id based on user input
+
+    Args:
+        Ctx: The click context
+
+    Returns:
+        None
+
+    Raises:
+        ClickException if bad environment number
+    """
+    env_list = get_environment_list()
+    index = 1
+    click.echo("Available Environments:")
+    for item in env_list:
+        click.echo(f"    {index}) {item.get('name')}")
+        index += 1
+    user_choice = click.prompt("Select environment", type=int)
+    try:
+        value = env_list[user_choice - 1]
+    except IndexError:
+        raise click.ClickException("Environment number chosen is not valid")
+    click.echo(f"Active environment is now {value.get('name')}")
+    save_config_value("current_environment", json.dumps(value))
+
+
+@environment_cmd.command()
+@click.pass_context
+def list(ctx):
+    """
+    Command to list all possible environments
+
+    Args:
+        Ctx: The click context
+
+    Returns:
+        None
+    """
+    env_list = get_environment_list()
+    current_env_id = json.loads(get_config_value("current_environment")).get("id")
+    for item in env_list:
+        name = item.get("name")
+        active = "  "
+        if current_env_id == item.get("id"):
+            active = "* "
+
+        click.echo(f"{active}{name}")
--- a/cli/functionary/environment.py
+++ b/cli/functionary/environment.py
@@ -40,7 +40,7 @@ def _get_environment_list():
 @click.pass_context
 def environment_cmd(ctx):
     """
-    Click group to associate all environment related commands with
+    List or set environments
     """
     pass
 
@@ -55,7 +55,7 @@ def set(ctx):
     index = 1
     click.echo("Available Environments:")
     for item in env_list:
-        click.echo(f"    {index}) {item.get('name')} - {item.get('team')}")
+        click.echo(f"    {index}) {item.get('team')} - {item.get('name')}")
         index += 1
     user_choice = click.prompt("Select environment", type=int)
     try:
@@ -86,4 +86,4 @@ def list(ctx):
         if current_env_id == item.get("id"):
             active = "* "
 
-        click.echo(f"{active}{name} - {team}")
+        click.echo(f"{active}{team} - {name}")
--- a/cli/functionary/environment.py
+++ b/cli/functionary/environment.py
@@ -6,23 +6,19 @@ import requests
 from .config import get_config_value, save_config_value
 
 
-def get_environment_list():
+def _get_environment_list():
     """
     Helper function to get the environment list from host
-
-    Args:
-        None
-
-    Returns:
-        None
-
-    Raises:
-        ClickException if bad get request
     """
     token = get_config_value("token")
     teams_url = get_config_value("host") + "/api/v1/teams"
     header = {"Authorization": f"Token {token}"}
-    response = requests.get(teams_url, headers=header)
+    try:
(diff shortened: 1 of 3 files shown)


cli/functionary/environment.py
```diff
@@ -0,0 +1,98 @@
+import json
+
+import click
+import requests
+
+from .config import get_config_value, save_config_value
+
+
+def get_environment_list():
```
scott-taubman, line 9: For these sorts of helper functions that are only intended for use internally in the same file, you can name them with a prepended underscore to designate them as such and then the docstring can just be the the brief description.  So like:

```
def _get_environments():
    """Helper function to get the list of available environment"""
```

There's nothing wrong with having the more verbose docstring, just that it's not required in this situation.  If you decide to keep it, your `Returns` is incorrect, as this appears to return a list of environments.

cli/functionary/environment.py
```diff
@@ -0,0 +1,98 @@
...
+    Returns:
+        None
+
+    Raises:
+        ClickException if bad get request
+    """
+    token = get_config_value("token")
+    teams_url = get_config_value("host") + "/api/v1/teams"
+    header = {"Authorization": f"Token {token}"}
+    response = requests.get(teams_url, headers=header)
```
scott-taubman, line 25: This can throw an exception.  For instance, if the host is unavailable.  You'll want to wrap it in a try / except block so that it can raise the more friendly ClickException rather than throwing a stack trace.

cli/functionary/environment.py
```diff
@@ -0,0 +1,98 @@
...
+
+    if response.ok:
+        data = json.loads(response.text).get("results")
+
+        env_list = []
+        for team in data:
+            for env_set in team.get("environments"):
+                env_list.append(env_set)
+        return env_list
```
scott-taubman, line 34: You'll want to capture the team name along with the environment in some way here.  Without it, you lose some important context when you display the list.  Here's what my output looks like when I run `functionary environment set`:

```shell
$ functionary environment set
Available Environments:
    1) default
    2) dev
    3) prod
    4) default
Select environment:
```

Without the team name, there's no way to differentiate the environments that happen to have the same name.

cli/functionary/environment.py
```diff
@@ -0,0 +1,98 @@
...
+    else:
+        raise click.ClickException(
+            f"Failed to get environment list: {response.status_code}\n"
+            f"Response: {response.text}"
+        )
+
+
+@click.group("environment")
+@click.pass_context
+def environment_cmd(ctx):
```
scott-taubman, line 44: You'll want a basic docstring here so that the usage has a nice summary.  It's used to fill in the command description when you just run `functionary` on its own:

```shell
$ functionary
Usage: functionary [OPTIONS] COMMAND [ARGS]...

Options:
  --help  Show this message and exit.

Commands:
  environment  # docstring summary will populate text here
  login        Login to Functionary.
  package
```

cli/functionary/environment.py
```diff
@@ -0,0 +1,98 @@
...
+    pass
+
+
+@environment_cmd.command()
+@click.pass_context
+def set(ctx):
+    """
+    Command to set the environment id based on user input
```
scott-taubman, line 52: You'll want this summary line to read as something that will be useful to someone running the `functionary environment` command, since the help text will be populated from here.  Something simple like:

> Set the active environment

cli/functionary/environment.py
```diff
@@ -0,0 +1,98 @@
...
+        raise click.ClickException("Environment number chosen is not valid")
+    click.echo(f"Active environment is now {value.get('name')}")
+    save_config_value("current_environment", json.dumps(value))
+
+
+@environment_cmd.command()
+@click.pass_context
+def list(ctx):
+    """
+    Command to list all possible environments
```
scott-taubman, line 82: Let's go with "List available environments"

cli/functionary/environment.py
```diff
@@ -0,0 +1,98 @@
...
+    pass
+
+
+@environment_cmd.command()
+@click.pass_context
+def set(ctx):
+    """
+    Command to set the environment id based on user input
+
+    Args:
```
scott-taubman, line 54: I didn't think about this before so if I previously provided confusing guidance I apologize...

The docstring here is what will be used to populate the output of `functionary environment set --help`.  Given that, you'll want to write the docstrings for the command functions from the perspective of being helpful user documentation, as opposed to our usual google style docstring.

cli/functionary/environment.py
```diff
@@ -0,0 +1,98 @@
...
+    for item in env_list:
+        click.echo(f"    {index}) {item.get('name')}")
+        index += 1
+    user_choice = click.prompt("Select environment", type=int)
+    try:
+        value = env_list[user_choice - 1]
+    except IndexError:
+        raise click.ClickException("Environment number chosen is not valid")
+    click.echo(f"Active environment is now {value.get('name')}")
+    save_config_value("current_environment", json.dumps(value))
```
scott-taubman, line 75: I don't think there's a reason to store the entire environment object.  I believe all you'll need is the id.  You can store that as a string and not bother with all of the json conversion when reading / writing the config.

cli/functionary/environment.py
```diff
@@ -0,0 +1,98 @@
...
+    Command to list all possible environments
+
+    Args:
+        Ctx: The click context
+
+    Returns:
+        None
+    """
+    env_list = get_environment_list()
+    current_env_id = json.loads(get_config_value("current_environment")).get("id")
```
scott-taubman, line 91: It should be possible to list the environments even if you've not yet set and active one.  As it currently stands that's not the case because this will immediately exit.

scott-taubman (changes requested):

cli/functionary/environment.py
```diff
@@ -42,57 +39,51 @@ def get_environment_list():
...
-        ClickException if bad environment number
+    Set the active environment
     """
-    env_list = get_environment_list()
+    env_list = _get_environment_list()
     index = 1
     click.echo("Available Environments:")
     for item in env_list:
-        click.echo(f"    {index}) {item.get('name')}")
+        click.echo(f"    {index}) {item.get('name')} - {item.get('team')}")
```
scott-taubman, line 58: This displays more cleanly with the team name listed first, then the name.  It ends up looking something like:

```
Available Environments:
    1) beergarden - default
    2) beergarden - dev
    3) beergarden - prod
    4) foxhound - default
Select environment:
```

Which just looks cleaner with the team names nicely lined up.

scott-taubman (changes requested):

scott-taubman (approved):

PR #42, opened 2022-08-18
#41 - Filter team endpoint based on user roles

### LINKED ISSUE #41: Filter /teams endpoint by the teams and environments a user is a member of
The `/api/v1/teams` endpoint should only show teams and environments that a user is a member of.  If the user has a role on the team, they should see the team and all of its environments.  If they have a role on an environment, they should see the team and only those environments on which they have an assigned role.

scott-taubman:
Closes #41 

This PR adds filtering on the `/api/v1/teams` endpoint to only show the teams and environments to which a user belongs.  The behavior is now this:

* If the user is a superuser, all teams and environments will be shown.
* If the user has a role on a team, that team and all of its environments will be shown.
* If the user has a role on an environment, that environment's team will shown, but only the environments that the user has a role on will be shown for that team.

There's also a fix to the unique constraint on user roles.  The intent was that a user should have only a single role assigned per team or environment.  That is now properly enforced.

### DIFF
--- a/functionary/core/api/v1/serializers/team.py
+++ b/functionary/core/api/v1/serializers/team.py
@@ -15,8 +15,30 @@ class TeamEnvironmentSerializer(serializers.ModelSerializer):
 class TeamSerializer(serializers.ModelSerializer):
     """Basic serializer for the Team model"""
 
-    environments = TeamEnvironmentSerializer(many=True)
+    environments = serializers.SerializerMethodField()
 
     class Meta:
         model = Team
         fields = ["id", "name", "environments"]
+
+    def get_environments(self, team):
+        """In the context of a request, filters the list of environments down to
+        those of the requesting user."""
+
+        request = self.context.get("request")
+
+        # If we are using the serializer outside of the context of a request, do not
+        # filter the environments.
+        #
+        # In the context of a request, if the user is a superuser or has a role on the
+        # team itself we do not need to filter the environments.
+        if (
+            request is None
+            or request.user.is_superuser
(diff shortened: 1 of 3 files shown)


kvig (approved):
Worked as described for me.

Two things I noticed, the error message when the team/user/role constraint is violated in the admin could be worded better, not sure if that's an easy thing to do or if it's something to be addressed later for an admin ui. 
Also I wanted to switch my default team environment and had to unset the default for the team so it didn't have one, then set the new one. We probably need add an operation on the Team in the future to allow this change and ensure consistency.

scott-taubman:
> Two things I noticed, the error message when the team/user/role constraint is violated in the admin could be worded better, not sure if that's an easy thing to do or if it's something to be addressed later for an admin ui. Also I wanted to switch my default team environment and had to unset the default for the team so it didn't have one, then set the new one. We probably need add an operation on the Team in the future to allow this change and ensure consistency.

I agree with both of these observations.

* I don't expect the django admin UI to be a thing that's user facing.  In my experience on past projects, you run into the limits of what the admin UI does out of the box pretty quickly.  Especially as your models and related functionality start to grow in complexity.  You can actually customize the admin UI quite a lot, but the amount of effort to do that is not worlds apart from what it would take to just build some administrative pages into your actual application UI in many cases.  So for now I'm just viewing the admin UI as a convenience tool for us to setup data in our dev environments.

* Yes, there needs to be some helper for setting the default environment.  I recognized this too, but I punted on implementing such a thing for now because there's a few ways to go about it and I figure which route to go might be more obvious once we have a context where it'll actually be used.

PR #46, opened 2022-08-23
#43 - Centralize request get/post

### LINKED ISSUE #43: Centralize request making and error handling
There is a bit of setup to do and error handling to consider every time the CLI code needs to make an API call.  For example, if you look at the logic for [getting the environment list](https://github.com/beer-garden/functionary/blob/main/cli/functionary/environment.py#L16) and [publishing a package](https://github.com/beer-garden/functionary/blob/main/cli/functionary/package.py#L95), you'll note that in each case the following has to happen:

* Get the current value of the "host" setting
* Get the current value of the "token" setting
* Construct the Authorization header
* Handle requests exceptions
* Handle various error response codes

As we expand the available commands, building the X-Environment-Id will fall into this category as well.

What we should do is centralize all of this logic in a module (let's say, `client.py` for now).  Then, instead of request code sprinkled throughout the various CLI functions looking like this:

```python
    token = get_config_value("token")
    teams_url = get_config_value("host") + "/api/v1/teams"
    header = {"Authorization": f"Token {token}"}
    try:
        response = requests.get(teams_url, headers=header)
    except requests.ConnectionError:
        raise click.ClickException("Could not connect to host")
    except requests.Timeout:
        raise click.ClickException("Timeout occured while trying to connect to host")

    if response.ok:
        # stuff happens here
    else:
        raise ClickException("...")
```

It could instead look like:

```python
    teams = client.get("teams")

    # You'll only get here if the request was successful, so you just carry on assuming teams is usable
    for team in teams:
       # stuff happens
```

The `client.get()` call takes care of all of the fiddly bits mentioned earlier:

* Constructs the Authorization and X-Environment-Id headers
* Handles exceptions
* Handles non 2xx responses.  Most specifically, telling the user to login if a 401 is received and that access was denied if a 403 is received.

Note that while the example here is for a `get`, we'll need a `post` as well.

catattack05:
Closes #43 

Added a new file called client.py to hold one post/get command for the whole CLI, then refactored the other files as needed


cli/functionary/client.py
```diff
@@ -0,0 +1,98 @@
...
+        raise click.ClickException("Timeout occured while trying to connect to host")
+
+    if response.ok:
+        data = json.loads(response.text).get("results")
+        return data
+    else:
+        if response.status_code == 401:
+            raise click.ClickException("Unauthorized request, try logging in again.")
+        elif response.status_code == 403:
+            raise click.ClickException("Request forbidden")
```
scott-taubman, line 41: Let's try to be more clear in the messages returned.  Maybe:

"You do not have access to perform this action"

cli/functionary/client.py
```diff
@@ -0,0 +1,98 @@
...
+        raise click.ClickException("Could not connect to host")
+    except requests.Timeout:
+        raise click.ClickException("Timeout occured while trying to connect to host")
+
+    if response.ok:
+        data = json.loads(response.text).get("results")
+        return data
+    else:
+        if response.status_code == 401:
+            raise click.ClickException("Unauthorized request, try logging in again.")
```
scott-taubman, line 39: "Authentication failed. Please login and try again."

cli/functionary/client.py
```diff
@@ -0,0 +1,98 @@
...
+        Endpoint: the name of the endpoint to get data from
+
+    Returns:
+        Upload_Response: Response from endpoint as Python list/dict
+
+    Raises:
+        ClickException if failed connection, timeout, or could not retrieve endpoint
+    """
+    token = get_config_value("token")
+    url = get_config_value("host") + f"/api/v1/{endpoint}"
```
scott-taubman, line 63: A lot of stuff is repeated between the `get` and `post` functions.  You'll want to refactor things in such a way that those common things happen in one place and the `get` and `post` just leverage that shared functionality.

cli/functionary/client.py
```diff
@@ -0,0 +1,98 @@
...
+    upload_response = None
+    try:
+        if endpoint == "publish":
```
scott-taubman, line 66: You don't need this check.  Just set the environment header if you can.  Even if an endpoint doesn't need it, including it won't hurt anything.  Which probably brings up another condition that you'll want to account for: The environment_id is not configured (the user hasn't run `functionary environment set` yet).  Most endpoints will require that with the `/teams` endpoint maybe being the only exception.  Raising an error telling the user they need to run `functionary environment set` when appropriate would be good.

cli/functionary/client.py
```diff
@@ -0,0 +1,98 @@
...
+    if upload_response is not None:
+        if upload_response.ok:
+            return json.loads(upload_response.text)
+        elif upload_response.status_code == 401:
+            raise click.ClickException("Unauthorized request, try logging in again.")
+        elif upload_response.status_code == 403:
+            raise click.ClickException("Request forbidden")
+        else:
+            raise click.ClickException(
+                f"Failed to build image: {upload_response.status_code}\n"
```
scott-taubman, line 94: This message is kind of a holdover that's no longer appropriate, but more importantly there shouldn't be any endpoint specific messaging in these `get` and `post` functions.

cli/functionary/client.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Raises:
+        ClickException if failed connection, timeout, or could not retrieve endpoint
+    """
+    token = get_config_value("token")
+    url = get_config_value("host") + f"/api/v1/{endpoint}"
+    upload_response = None
+    try:
+        if endpoint == "publish":
+            environment_id = json.loads(get_config_value("current_environment")).get(
```
scott-taubman, line 67: This `json.loads` things was actually an error that slipped through before, which you have now copied here.  It's since been fixed, but basically this should just be:

```python
environment_id  get_config_value("current_environment_id")
```

scott-taubman (changes requested):

cli/functionary/client.py
```diff
@@ -0,0 +1,104 @@
...
+
+    Raises:
+        ClickException: Raised if cannot connect to host, permission issue
+        exists, user has not set a required field, or other request failure
+
+    """
+    host = get_config_value("host")
+    url = host + f"/api/v1/{endpoint}"
+    try:
+        if endpoint == "api-token-auth" and request_type == "post":
```
scott-taubman, line 65: There shouldn't be any checks for specific endpoints within this code.  There are two possibilities here:

* The call to the api-token-auth endpoint has some completely unique need that nothing else going through this code path might share.  If that is the case, it would make more sense to just have the spot where api-token-auth is called not use this function and instead just call `requests.post` directly.
* Or, the more likely scenario, the api-token-auth call is indicative of a more general use case that this function could be reworked to accommodate.

Let's say it's that second one and that the general use case is calls to endpoints where authorization is not required.  That is, calls that you can make or are expected to make without having first logged in.  The act of logging in itself is the situation you're dealing with here, but there could conceivably be others.

You could handle this by having this (and the `get` and `post` functions that call it) take an argument `authorization` which defaults to `True`, but in the login scenario you would set to `False`.  Then that would drive whether or not you set the authorization header here.

Having said all that, the first route to try is just always sending the authorization token if you have it and not sending it if you don't.  In the login scenario, if the user already has an authorization token and is trying to login (possibly under a different account), it's okay to send that old token because it's just going to be ignored by that endpoint anyway.  For other endpoints, if the token was required and you don't send it, the response from the API is going to tell you that.

cli/functionary/client.py
```diff
@@ -0,0 +1,104 @@
...
+            else:
+                try:
+                    environment_id = get_config_value("current_environment_id")
+                except click.ClickException:
+                    raise click.ClickException(
+                        "Please set an active environment id using 'environment set'"
+                    )
+                headers = {
+                    "Authorization": f"Token {token}",
+                    "X-Environment-ID": f"{environment_id}",
```
scott-taubman, line 81: Similar comment to the one about the authorization token.  The API response will inform you failed to send the `X-Environment-Id` header in a situation that required it.  So you again should be able to just set the header if you have a value, not set it if you don't, do the request, and then if you get the 400 response telling you that the header is missing, raise your `ClickException("Please set an active environment id using 'environment set'")`.

The idea is to leverage the logic built into the API as best you can, rather than having to write a bunch of endpoint specific checks here in the CLI.

cli/functionary/environment.py
```diff
@@ -1,39 +1,21 @@
...
-    try:
-        response = requests.get(teams_url, headers=header)
-    except requests.ConnectionError:
-        raise click.ClickException("Could not connect to host")
-    except requests.Timeout:
-        raise click.ClickException("Timeout occured while trying to connect to host")
-
-    if response.ok:
-        data = json.loads(response.text).get("results")
+    data = get("teams")
```
scott-taubman, line 11: You may as well just call this `teams` rather than `data`.  In this particular case we never get too far away from this declaration that one would lose track of what this is, but as a general practice you want to have descriptive names for things so that someone reading a bit of code that uses this would still have some clue what it is even removed from the context of this assignment.

cli/functionary/tokens.py
```diff
@@ -1,25 +1,9 @@
...
-            )
-    except requests.ConnectionError:
-        raise click.ClickException(f"Unable to connect to {login_url}.")
-    except requests.Timeout:
-        raise click.ClickException("Timeout occurred waiting for login")
+def login(user: str, password: str):
+    login_response = post(
+        "api-token-auth", data={"username": user, "password": password}
+    )
+    token = login_response.get("token")
```
scott-taubman, line 8: The most super minor thing, but I notice this pattern a lot of setting a variable, only to return it on the very next line.  There's nothing wrong with that really, but you can just return things directly:

```python
return login_response.get("token")
```

cli/functionary/package.py
```diff
@@ -72,41 +72,16 @@ def publish(ctx, path):
...
-    elif upload_response.status_code == 401:
-        raise click.ClickException("\n\nUnauthorized request, try logging in again.")
-    else:
-        raise click.ClickException(
-            f"Failed to build image: {upload_response.status_code}\n"
-            f"\tResponse: {upload_response.text}"
-        )
+    post("publish", files={"package_contents": upload_file})
+
+    click.echo("Publish succeded")
```
scott-taubman, line 87: It might actually be useful to output the build id here.  That way if / when we add a command for viewing build status, the user has the id they need to look it up.

cli/functionary/package.py
```diff
@@ -72,41 +72,16 @@ def publish(ctx, path):
...
     """
-    token = get_config_value("token")
     host = get_config_value("host")
 
     full_path = pathlib.Path(path).resolve()
     tarfile_name = full_path.joinpath(f"{full_path.name}.tar.gz")
     with tarfile.open(str(tarfile_name), "w:gz") as tar:
         tar.add(str(full_path), arcname="")
 
+    upload_file = open(tarfile_name, "rb")
```
scott-taubman, line 82: This looks like it was true before, but this file gets opened and then isn't closed anywhere.  Seems like something to address.  You can use `open` in a [context manager](https://book.pythontips.com/en/latest/context_managers.html) to fix this.

cli/functionary/client.py
```diff
@@ -0,0 +1,104 @@
...
+                }
+                if request_type == "post":
+                    response = requests.post(
```
scott-taubman, line 84: This one is kind of take it or leave it, but there is actually a [`requests.request`](https://requests.readthedocs.io/en/latest/api/#requests.request) that you takes the HTTP method as a parameter.  You could leverage that to avoid having to do this `request_type` check to figure out whether you should call `get` or `post`.

scott-taubman (changes requested):

cli/functionary/client.py
```diff
@@ -61,31 +59,26 @@ def _send_request(endpoint, request_type, post_data=None, post_files=None):
...
-                    response = requests.post(
-                        url, headers=headers, data=post_data, files=post_files
-                    )
-                else:
-                    response = requests.get(url, headers=headers)
+            headers["Authorization"] = f"Token {token}"
+        except click.ClickException:
+            pass
+
+        try:
```
scott-taubman, line 70: This whole `try` and then `pass` thing is a little awkward.  Instead, what I think is appropriate here is to modify `get_config_value` so that it takes an optional `raise_exception` parameter that defaults to `True`.  If `raise_exception` is `False`, it would just return the `None` when a value can't be found.  Then here and for the token you could just do:

```python
if (environment_id := get_config_value("current_environment_id", raise_exception=False)) is not None:
    headers["X-Environment-ID"] = environment_id
```

cli/functionary/client.py
```diff
@@ -94,6 +87,10 @@ def _send_request(endpoint, request_type, post_data=None, post_files=None):
 
     if response.ok:
         return response
+    elif response.status_code == 400:
+        raise click.ClickException(
+            "Please set an active environment id using 'environment set'"
```
scott-taubman, line 92: For right now I think we can just go with this and assume that a 400 is because the header is missing.  In reality, a 400 could represent a whole host of things.  Since the errors returned don't currently include a good, parsable error code (you get the HTTP status code of 400, and then some human friendly message), there's not a good way to discern why the 400 was returned here.  So keep this as is for now and at some point when the API returns errors codes we can come back and update this handling.

Maybe put a comment here explaining that:

```python
# TODO: Add checking of error codes in addition to status once they are added to the API
```

cli/functionary/package.py
```diff
@@ -82,6 +82,7 @@ def publish(ctx, path):
     upload_file = open(tarfile_name, "rb")
     click.echo(f"Publishing {str(tarfile_name)} package to {host}")
 
-    post("publish", files={"package_contents": upload_file})
-
-    click.echo("Publish succeded")
+    response = post("publish", files={"package_contents": upload_file})
+    upload_file.close()
```
scott-taubman, line 86: Look at using a context manager here as suggested earlier.  The reason why that's preferable here is that the file will still get closed even if your `post` call raises an exception.  From [the docs](https://docs.python.org/3/library/io.html#io.IOBase):

> IOBase is also a context manager and therefore supports the with statement. In this example, file is closed after the with statement’s suite is finished—even if an exception occurs:

As written here, your `close()` call would never be reached in the case of an exception.

scott-taubman (changes requested):

cli/functionary/package.py
```diff
@@ -72,41 +72,15 @@ def publish(ctx, path):
...
-    else:
-        raise click.ClickException(
-            f"Failed to build image: {upload_response.status_code}\n"
-            f"\tResponse: {upload_response.text}"
-        )
+    with open(tarfile_name, "rb") as upload_file:
+        click.echo(f"Publishing {str(tarfile_name)} package to {host}")
+        response = post("publish", files={"package_contents": upload_file})
+        id = response["id"]
+        click.echo(f"Publish {id} succeded")
```
scott-taubman, line 86: This wording doesn't make it totally clear what the `id` is that is being returned.  Also, "succeeded" is misspelled.  How about:

```python
click.echo(f"Package upload complete\nBuild id: {id}")
```

cli/functionary/config.py
```diff
@@ -52,7 +52,10 @@ def get_config_value(key):
         config_file = Path.home() / ".functionary" / "config"
         value = get_key(config_file, key)
         if value is None:
-            raise click.ClickException(f"Could not find value for {key}")
+            if raise_exception is not None:
```
scott-taubman, line 55: `raise_exception` is a boolean, so you actually want to be checking the value, or at least the truthiness of it.  As written, if someone passes in `True`, it won't actually raise an exception, because per this check, `True` is in fact, `not None`.

Also worth mentioning here, is that you should look at the other uses of `get_config_value` now that you've made this change with `raise_exception`.  Specifically, there's a use in `environment.py` that is has a try / except block but no longer needs it because this will now safely return `None`.

scott-taubman (changes requested):

scott-taubman (approved):

PR #48, opened 2022-08-25
#45 - Task model and view

### LINKED ISSUE #45: Create Task model and views
We need a Task model and corresponding ViewSet for creating a viewing tasks (individual executions of a function).

scott-taubman:
Closes #45

This PR adds the Task model and ViewSet.  The logic that was previously in the `ExecuteView` has been moved to the various appropriate places for Task (validation on the model, request schema to a serializer, etc).

## Testing Instructions

To create a task, you now POST to the /api/v1/tasks endpoint.  There are two ways to task:

* supplying a function ID via the "function" parameter.
* supplying a "function_name" and "package_name" parameter

These two options for how to task are documented in the request input schema for the endpoint on the swagger docs.

The api response for a successfully created task is simply the task id.  If validation of the parameters against the function schema fails, a 400 response is returned with an appropriate error message.

There are also corresponding list and retrieve endpoints (GET /api/v1/tasks and /api/v1/tasks/{id}) for viewing tasks.

## Other Notes
* The header definition that was in `EnvironmentViewMixin.header_parameters` has been moved to `core.api.HEADER_PARAMETERS`.  This was done because `TaskViewSet` is built on the `EnvironmentGenericViewSet` and not one of the classes that already defined those header parameters.  It seemed awkward to import `EnvironmentViewMixin` just to use it's `header_parameters` property.  Instead, I opted to do this slight refactor.
* The changes here don't actually result in tasking getting executed.  This just creates the entry in the database.  The actual tasking message getting sent to the runner will happen in a separate PR.


functionary/core/models/task.py
```diff
@@ -0,0 +1,93 @@
...
+        """Validate that the parameters conform to the function's schema"""
+        try:
+            jsonschema.validate(
+                instance=self.parameters,
+                schema=self.function.schema,
+            )
+        except jsonschema.exceptions.ValidationError as exc:
+            raise ValidationError(exc.message)
+
+    def clean(self):
```
scott-taubman, line 90: This `clean` pattern isn't the most intuitive, so just in case, here's a [reference](https://docs.djangoproject.com/en/4.0/ref/models/instances/#django.db.models.Model.clean).

functionary/core/models/task.py
```diff
@@ -0,0 +1,93 @@
...
+            models.Index(
+                fields=["environment", "updated_at"], name="task_environment_updated_at"
+            ),
+        ]
+
+    def _clean_environment(self):
+        """Ensures that the environment is correctly set to that of the function"""
+        if self.environment is None:
+            self.environment = self.function.package.environment
+        elif self.environment != self.function.package.environment:
```
kvig, line 75: Is this the intended logic? As it is, you have to keep guessing the environment until you find the right one for the function.

kvig (changes requested):

functionary/core/models/task.py
scott-taubman: I think this is correct, but I'm not especially in love with it.   Or at least, I'm not in love with the way model instance validation is handled via the separate `clean` call and isn't integrated into the save process.

Regardless, the intention is that you should not necessarily have to explicitly set the environment.  If you leave environment as `None`, then when you call `clean()` the environment will get properly set according to the environment for the function.  But, **if you do set the environment** then it must be the environment of the function, otherwise bad things would happen, thus the `ValidationError`.

functionary/core/models/task.py
```diff
@@ -0,0 +1,93 @@
...
+
+import jsonschema
+from django.conf import settings
+from django.core.exceptions import ValidationError
+from django.db import models
+
+from core.models import Environment, Function
+
+
+class Task(models.Model):
```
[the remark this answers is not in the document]
scott-taubman: Yup.  Added.

kvig (approved):

PR #49, opened 2022-08-29
Issue 47: Adding buildstatus and table output

### LINKED ISSUE #47: View build status in CLI
We need a general means of nicely displaying data that is queried by the CLI.  One example is the package build status.  When a user calls:

```shell
functionary package publish /path/to/package
```

They [will soon] receive back the id of the build for that package.  We want them to then be able to query for that status of that build via the CLI.  Something like:

```shell
# show status of all builds
functionary package buildstatus

# get status of a specific build
functionary package buildstatus --id somebuildid
```

The two examples above would hit the `/api/v1/builds` and `/api/v1/builds/{id}` endpoints.

This build status example is one of many cases like this we are going to have, so the important part of this issue is figuring out how to nicely display the results from an API response.

NOTE: The actual data that comes back from the build endpoint isn't super great at identifying what the build is about (you get uuids back instead of friendly names for things).  For now, just go with it and assume that we'll return friendlier values for those fields in the future.

catattack05:
Closes #47 

I have added a buildstatus command into package.py. This uses the rich library (which you'd need to install with pip install rich) to turn API data into a nice table. I've written a helper function called _format_results to turn a list of dictionaries into a Rich table. That function currently lives in package.py but it could live somewhere else if we'd rather.

Also, I changed environment list to have the active environment marked in red. Could change that color if needed.

### DIFF
--- a/cli/functionary/package.py
+++ b/cli/functionary/package.py
@@ -4,8 +4,10 @@ import tarfile
 
 import click
 import yaml
+from rich.console import Console
+from rich.table import Table
 
-from .client import post
+from .client import get, post
 from .config import get_config_value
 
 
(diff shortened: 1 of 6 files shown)


cli/functionary/package.py
```diff
@@ -84,3 +86,43 @@ def publish(ctx, path):
...
+    """
+    table = Table(title=title, width=170)
+    console = Console()
+    first_row = True
+    for item in results:
+        list = []
+        for key in item:
+            if first_row:
+                table.add_column(key.capitalize())
+            list.append(str(item[key]))
```
scott-taubman, line 125: As written, this ends up stringifying `None` values, making them show up as "None" rather than a blank cell, which is not ideal.

You can avoid that by doing this:

```python
list.append(str(item[key]) if item[key] else None)
```

Doing that leads to you resolving `item[key]` multiple times, which we should also avoid.  You can actually get the key and value at the same time at the top of your for loop by doing:

```python
for key, value in item.items():
```

Then when you get down here, you have the more sensible looking:

```python
list.append(str(value) if value else None)
```

cli/functionary/package.py
```diff
@@ -84,3 +86,43 @@ def publish(ctx, path):
...
+
+    Returns:
+        None
+    """
+    table = Table(title=title, width=170)
+    console = Console()
+    first_row = True
+    for item in results:
+        list = []
+        for key in item:
```
scott-taubman, line 122: I think it makes sense to just universally exclude the "environment" field.  It's always going to show the same value (the id of the currently set environment), which makes it not entirely useful.  It mostly just ends up taking space.  You can check the `key` and just call `continue` if it's "environment".  Or better yet, maybe create a constant called `EXCLUDED_FIELDS` and have your check be:

```python
if key in EXCLUDED_FIELDS:
    continue
```

scott-taubman (changes requested):

scott-taubman (approved):

PR #51, opened 2022-08-31
Issue 50: Add way to view packages and their functions

### LINKED ISSUE #50: List packages and their associated functions using Rich tables
We want to add a package list command to the CLI that will print out a series of tables using Rich. There will be one table per package, and within the table will be all the functions associated with that package + other relevant data

catattack05:
Closes #50 

Here's one possible way we could go about viewing packages + functions. Right now in the basic view command, every function gets its own table with all its different functions and some relevant information (I could add things like schema back in if we wanted them).

Also, since package buildstatus has an option to query builds by id, I did add in an optional argument to query by package id so we just spit out one table. I'm not entirely sure if we want it, but I could see it being helpful if you just want information about one particular package.

### DIFF
--- a/cli/functionary/package.py
+++ b/cli/functionary/package.py
@@ -96,7 +96,61 @@ def buildstatus(ctx, id):
     """
     if id:
         results = get(f"builds/{id}")
-        format_results([results], title=f"Build: {id}")
+        format_results([results], title=f"Build: {id}", excluded_fields=["environment"])
     else:
         results = get("builds")
-        format_results(results, title="Build Status")
+        format_results(results, title="Build Status", excluded_fields=["environment"])
+
+
+@package_cmd.command()
+@click.pass_context
+@click.option("--id", help="view a specific package with the given id")
+def view(ctx, id):
+    """
+    View all current packages and their functions
+    """
+    packages = get("packages")
+    functions = get("functions")
+    if id:
+        name = ""
+        description = ""
+        found = False
+        for package in packages:
+            if package["id"] == id:
+                name = package["name"]
+                description = package["description"]
+                found = True
+        if found:
+            associated_functions = []
+            for function in functions:
+                if function["package"] == id:
+                    function_dict = {}
+                    function_dict["Function Name"] = function["name"]
+                    function_dict["Display Name"] = function["display_name"]
+                    function_dict["Description"] = function["description"]
+                    associated_functions.append(function_dict)
+            format_results(
+                associated_functions,
+                title=f"{name}\n {description}",
+            )
+        else:
+            click.echo(f"Package with id {id} not found")
+    else:
+        for i in range(0, len(packages)):
+            package = packages[i]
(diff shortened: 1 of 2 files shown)


cli/functionary/package.py
```diff
@@ -96,7 +97,64 @@ def buildstatus(ctx, id):
...
+            associated_functions = _find_functions(id, functions)
+        else:
+            raise click.ClickException(f"Package with id {id} not found")
+    else:
+        for i in range(0, len(packages)):
+            package = packages[i]
+            name = package["name"]
+            id = package["id"]
+            description = package["description"]
+            associated_functions = _find_functions(id, functions)
```
scott-taubman, line 134: In terms of perceptible performance difference this isn't likely to change anything, but I will point out that your implementation here has you iterating over the full set of functions N times, where N is the number of packages.  It should be possible to do this in a way that requires you to iterate over the function list once and only once, regardless of how many packages there are.  I'll leave the details of that as an exercise for you.

cli/functionary/package.py
```diff
@@ -96,7 +97,64 @@ def buildstatus(ctx, id):
...
+
+    Returns:
+        associated_functions: A list of dictionaries with information about each
+        function
+    """
+    associated_functions = []
+    for function in functions:
+        if function["package"] == id:
+            function_dict = {}
+            function_dict["Function Name"] = function["name"]
```
scott-taubman, line 156: Let's try calling this "Function" rather than "Function Name".

cli/functionary/package.py
```diff
@@ -96,7 +97,64 @@ def buildstatus(ctx, id):
...
+    else:
+        for i in range(0, len(packages)):
+            package = packages[i]
+            name = package["name"]
+            id = package["id"]
+            description = package["description"]
+            associated_functions = _find_functions(id, functions)
+    title = Text(f"{name}", style="bold green")
+    title.append(f"\n{description}", style="green dim")
+    format_results(associated_functions, title=title)
```
scott-taubman, line 137: I'm pretty sure that as written this is only going to output the last package result rather than all of them.

cli/functionary/package.py
```diff
@@ -96,7 +97,37 @@ def buildstatus(ctx, id):
...
+            index_list = [function_dict]
+            functions_lookup[package_id] = index_list
+    for package in packages:
+        name = package["name"]
+        id = package["id"]
+        description = package["description"]
+        associated_functions = functions_lookup[id]
+        title = Text(f"{name}", style="bold blue")
+        title.append(f"\n{description}", style="blue dim")
+        format_results(associated_functions, title=title)
```
scott-taubman, line 133: A blank line between packages helps readability IMO.  Let's throw a `click.echo("\n")` after this `format_results` call.

scott-taubman (changes requested):

scott-taubman (approved):

PR #53, opened 2022-09-01
Issue 52: Make login independent get/post

### LINKED ISSUE #52: Make login independent of centralized get/post requests
Apparently centralizing get/post has broken the login, so we're going to take it back to functioning independently for now

catattack05:
Closes #52 

Apparently using the centralization for get/post broke login, so I reverted login back to what it was before.

### DIFF
--- a/cli/functionary/login.py
+++ b/cli/functionary/login.py
@@ -16,8 +16,7 @@ def login_cmd(ctx, user, password, host):
     Set the output of this command to the FUNCTIONARY_TOKEN environment variable
     for other functionary commands to use to communicate with the server.
     """
-    login_url = f"{host}/api/v1/api-token-auth"
     save_config_value("host", host)
-    token = login(login_url, user, password)
+    token = login(f"{host}/api/v1/api-token-auth", user, password)
     save_config_value("token", token)
     click.echo("Login successful!")
--- a/cli/functionary/login.py
+++ b/cli/functionary/login.py
@@ -17,6 +17,6 @@ def login_cmd(ctx, user, password, host):
     for other functionary commands to use to communicate with the server.
     """
     save_config_value("host", host)
-    token = login(f"{host}/api/v1/api-token-auth", user, password)
+    token = login(host, user, password)
     save_config_value("token", token)
     click.echo("Login successful!")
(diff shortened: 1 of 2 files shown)


cli/functionary/login.py
```diff
@@ -16,7 +16,8 @@ def login_cmd(ctx, user, password, host):
     Set the output of this command to the FUNCTIONARY_TOKEN environment variable
     for other functionary commands to use to communicate with the server.
     """
+    login_url = f"{host}/api/v1/api-token-auth"
     save_config_value("host", host)
-    token = login(user, password)
+    token = login(login_url, user, password)
```
scott-taubman, line 21: Just pass in the host here and have the login function construct the URL.

cli/functionary/tokens.py
```diff
@@ -1,8 +1,24 @@
...
+def login(login_url, user: str, password: str):
+    try:
+        login_response = requests.post(
+            f"{login_url}", data={"username": user, "password": password}
+        )
+        # check status code/message on return then exit
+        if login_response.ok:
+            token = json.loads(login_response.text).get("token")
+            return token
+        else:
```
scott-taubman, line 16: Let's add an `elif login_response.status_code == 400` and have it raise with a more friendly message, like "Unable to login due to an invalid username or password."

This else block is fine for any other (non-400) responses.

scott-taubman (changes requested):

cli/functionary/tokens.py
```diff
@@ -1,8 +1,28 @@
...
+        login_response = requests.post(
+            f"{login_url}", data={"username": user, "password": password}
+        )
+        # check status code/message on return then exit
+        if login_response.ok:
+            token = json.loads(login_response.text).get("token")
+            return token
+        elif login_response.status_code == 400:
+            raise click.ClickException(
+                "Code 400: Unable to login due to an invalid username or password."
```
scott-taubman, line 18: You don't need the "Code 400" part here.  That's surfacing too much of the implementation detail to the user.  Just the "Unable to login..." bit is necessary.

cli/functionary/login.py
```diff
@@ -17,6 +17,6 @@ def login_cmd(ctx, user, password, host):
     for other functionary commands to use to communicate with the server.
     """
     save_config_value("host", host)
-    token = login(user, password)
+    token = login(f"{host}/api/v1/api-token-auth", user, password)
```
scott-taubman, line 20: What I meant here was just pass the host in.  So instead of taking a url, `login` will take a host parameter.  Then construct the url inside of the login function.

scott-taubman (changes requested):

scott-taubman (approved):

PR #54, opened 2022-09-07
Create Functionary UI

kvig:
This PR creates a multipage UI using the Django templates. It supports views for all the object types and allows for submitting a tasking request. 

All views and tasking requests are filtered out to only be associated with the selected environment shown under the username.


functionary/ui/views/functions.py
```diff
@@ -0,0 +1,76 @@
...
+        env_id = self.request.session["environment_id"]
+        return (
+            super()
+            .get_queryset()
+            .filter(package__environment__id=env_id)
+            .order_by("package__name", "name")
+        )
+
+
+class FunctionDetailView(DetailView):
```
scott-taubman, line 28: These detail views actually need the `get_queryset` stuff too.  Otherwise what happens is that if you go to a URL for a specific function, package, etc, you'll see it regardless of which environment you have active.  This is easily reproduce-able by just going to a detail page and then switching the environment.

Having said that, we clearly need a mixin or base class or something that handles all of this environment filtering for us so that it's not defined on every view.  If you want to just put a `TODO` for adding queryset filtering, that's fine.

functionary/ui/views/functions.py
```diff
@@ -0,0 +1,76 @@
...
+                )
+
+                # redirect to a new URL:
+                return HttpResponseRedirect("/ui/task_list/")
+        else:
+            return HttpResponseForbidden()
+
+    # if a GET (or any other method), create a blank form
+    else:
+        func = Function.objects.get(id=request.GET["function_id"])
```
scott-taubman, line 69: Two things here:

* I wonder if a class based view would make the division between `get` and `post` logic more cleanly separated.
* In any situation where you are directly interacting with a queryset, be mindful of the environment based filtering.  This presumably will go off and render the form for whatever function id you provide, regardless of whether that function is in the currently selected environment or not.

functionary/ui/views/packages.py
```diff
@@ -0,0 +1,23 @@
...
+        env_id = self.request.session["environment_id"]
+        return super().get_queryset().filter(environment__id=env_id)
+
+
+class PackageDetailView(DetailView):
+    model = Package
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["functions"] = Function.objects.filter(package=self.get_object())
```
scott-taubman, line 22: You can leverage the django relationships here to get the available functions through the package itself, rather than having to import the `Function` model:

```python
context["functions"] = self.get_object().function_set.all()
```

If you go back to the `Package` model and set `related_name="functions"` on the `package` field, you can do a slightly more natural version of that:

```python
context["functions"] = self.get_object().functions.all()
```

functionary/functionary/settings/unicorn_.py
```diff
@@ -0,0 +1,3 @@
+UNICORN = {
```
scott-taubman, line 1: This probably happens naturally once the unicorn stuff is moved into the ui app, but just a note that I'd create a `ui_.py` settings file and put this there.

functionary/ui/templates/core/function_detail.html
```diff
@@ -0,0 +1,108 @@
...
+
+<link rel="stylesheet" href="{% static 'css/bulma-accordion-2.0.1.min.css' %}"/>
+
+<script src="{% static 'js/bulma-accordion-2.0.1.min.js' %}" defer></script>
+<script src="{% static 'js/bulma-toast-2.4.1.min.js' %}" defer></script>
+
+<script>
+    window.addEventListener('load', (event) => { bulmaAccordion.attach(); });
+</script>
+<script>
```
scott-taubman, line 55: I made a similar comment elsewhere, but this stuff all goes away now right?

functionary/ui/templates/core/package_list.html
```diff
@@ -0,0 +1,38 @@
...
+                <div class="media">
+                    <div class="media-left">
+                        <span class="icon is-large">
+                            <i class="fa fa-2x fa-box-open"></i>
+                        </span>
+                    </div>
+                    <div class="media-content">
+                        <a href="{% url 'ui:package-detail' package.id %}">
+                            <p class="title is-4">
+                                {% if package.display_name%}
```
scott-taubman, line 20: Just because it'll drive me nuts, I'm going to make a note here that we should at some point look into formatters that are compatible with django templating.  I don't know that there's a super obvious solution out there, but it'd save us from little inconsistencies, like the lack of a space before the closing `%}` here.

functionary/ui/forms/forms.py
```diff
@@ -0,0 +1,52 @@
...
+                "help_text": value.get("description", None),
+            }
+            field_class: Field = field_mapping.get(value["type"], None)
+
+            if not field_class:
+                raise ValueError(f"Unknown field type for {param}: {value['type']}")
+
+            field = field_class(**kwargs)
+            field.widget.attrs.update({"class": "input"})
+            # setattr(field, 'field_type', value['type'])
```
scott-taubman, line 51: Let's remove this commented out line.  If it was left in intentionally to remind you about something for later, use a `TODO` comment instead.

scott-taubman (changes requested):

functionary/ui/views/functions.py
```diff
@@ -0,0 +1,76 @@
...
+    func: Function = None
+    form: FunctionForm = None
+
+    if request.method == "POST":
+        env: Environment = Environment.objects.get(
+            id=request.session.get("environment_id")
+        )
+        permissions = request.user.environment_permissions(env, inherited=True)
+
+        if request.user.is_superuser or Permission.TASK_CREATE.value in permissions:
```
scott-taubman, line 48: I think you can avoid the `environment_permissions` call above and possibly just change this to be:

```python
if request.user.has_perm(Permission.TASK_CREATE, env):
```

scott-taubman: Long term, we should build helpers similar to those in the API views so that the proper response is handled for you if this check fails.  Just a note for the future.

functionary/ui/views/functions.py
```diff
@@ -0,0 +1,76 @@
...
+            .order_by("package__name", "name")
+        )
+
+
+class FunctionDetailView(DetailView):
+    model = Function
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        form: FunctionForm = FunctionForm(self.get_object())
```
scott-taubman, line 33: This is just a curiosity more than anything, what's the driver behind the type declaration here?  It's kind of obvious from the assignment what type this is going to end up being and then there's never a point after this where form is reassigned such that type checking would come into play.  Was there a specific thing you were you going for here?

functionary/ui/views/functions.py
```diff
@@ -0,0 +1,76 @@
...
+        context["form"] = form.render("forms/function_detail.html")
+        return context
+
+
+def execute(request) -> HttpResponse:
+    func: Function = None
+    form: FunctionForm = None
```
scott-taubman, line 40: Another type declaration comment...

In this case, you're declaring the type, but then immediately setting the value to something other than that type.  If you had some some sort of linter / static checker here it would almost certainly yell at you.

For example:

image

I am a proponent of type hints, most especially in function / method signatures.  For variable declarations it kind of feels like a judgement call in terms of the value added.  In this case, you can either change this to be something like:

```python
form: Optional[FunctionForm] = None
```

Or just opt to drop the type declaration.

functionary/ui/views/teams.py
```diff
@@ -0,0 +1,20 @@
...
+class TeamListView(ListView):
+    model = Team
+
+
+class TeamDetailView(DetailView):
+    model = Team
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["environments"] = Environment.objects.filter(team=self.get_object())
```
scott-taubman, line 16: I made a similar comment elsewhere, but you can get at the Environment and TeamUserRole objects through the Team:

```python
team = self.get_object()
context["environments"] = team.environments.all()
context["users"] = [ user_role.user for user_role in team.user_roles.all() ]
```

functionary/ui/forms/forms.py
```diff
@@ -0,0 +1,52 @@
...
+    DecimalField,
+    Field,
+    FloatField,
+    Form,
+    IntegerField,
+    JSONField,
+    TimeField,
+)
+
+field_mapping: dict[str, Field] = {
```
scott-taubman, line 15: Last type declaration comment I'll make...

This is presumably intended to be:
```python
field_mapping: dict[str, Type[Field]] = {
```

functionary/ui/templates/base.html
```diff
@@ -0,0 +1,109 @@
...
+                // rendered, the "block menu_select" parts are substituted by the template
+                const dest_url = '{% block menu_select %}{% endblock menu_select %}'.trim()
+                const menuItem = document.querySelector(`.menu-item > a[href$='${dest_url}'`);
+                menuItem.classList.add('is-active')
+            });
+
+            XMLHttpRequest.prototype._send = XMLHttpRequest.prototype.send;
+
+            //override send to always add the X-Environment-ID header for XHR's
+            XMLHttpRequest.prototype.send = function () {
```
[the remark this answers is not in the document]
kvig: Should be able to go, I forgot this was in there.

functionary/functionary/settings/django_.py
```diff
@@ -136,6 +138,7 @@
 # https://docs.djangoproject.com/en/4.0/howto/static-files/
 
 STATIC_URL = "static/"
+STATIC_ROOT = os.path.join(BASE_DIR.parent, "ui", "static")
```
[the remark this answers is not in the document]
kvig: I had added it because static stuff wasn't working at one point, but it appears to be fine now, removed.

functionary/functionary/settings/__init__.py
```diff
@@ -4,3 +4,4 @@
 from .django_ import *  # noqa
 from .logging_ import *  # noqa
 from .rest_framework_ import *  # noqa
+from .unicorn_ import * # noqa
```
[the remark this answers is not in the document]
kvig: I had manually added that line and didn't indent the noqa enough. Added the sort on save setting also.

functionary/core/auth/__init__.py
```diff
@@ -75,6 +75,7 @@ class Role(Enum):
 _DEVELOPER_PERMISSIONS = _READ_ONLY_PERMISSIONS + [
     Permission.PACKAGE_CREATE.value,
     Permission.PACKAGE_UPDATE.value,
+    Permission.TASK_CREATE.value,
```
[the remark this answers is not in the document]
kvig: Moved

functionary/functionary/urls.py
```diff
@@ -23,6 +23,7 @@
 from rest_framework.authtoken.views import obtain_auth_token
 
 urlpatterns = [
+    path("accounts/", include("django.contrib.auth.urls")),
```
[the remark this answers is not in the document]
kvig: I tried moving the url into UI, that didn't work. The APP_DIRS is True in the django_.py TEMPLATES section, so it's going to look for them in all the installed apps. UI is last so anything else will probably win.

I'll have to look into a more appropriate way of doing this. It also needs to block the UI from showing stuff if not logged in.

functionary/ui/views/environments.py
```diff
@@ -0,0 +1,33 @@
...
+class EnvironmentDetailView(DetailView):
+    model = Environment
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["packages"] = Package.objects.filter(environment=self.get_object())
+        return context
+
+
+@api_view(["POST"])
```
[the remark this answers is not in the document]
kvig: Fixed

functionary/functionary/settings/unicorn_.py
kvig: done

functionary/ui/views/functions.py
kvig: No real purpose, I had turned on type suggestions and clicked on some of them. But it didn't really make my editor view look better, so I removed them for your comments.

functionary/ui/views/teams.py
kvig: Updated.

functionary/ui/forms/forms.py
kvig: Removed

functionary/ui/templates/core/package_list.html
kvig: Added djlint to the project. Use the following to run it:
`djlint functionary/ui/templates --profile=django`

functionary/ui/forms/forms.py
```diff
@@ -0,0 +1,52 @@
...
+    "double": DecimalField,
+    "boolean": BooleanField,
+    "date": DateField,  # TODO set input_formats= on constructor if needed
+    "datetime": DateTimeField,  # TODO set input_formats= on constructor if needed
+    "time": TimeField,  # TODO set input_formats= on constructor if needed
+    "object": JSONField,
+}
+
+
+class FunctionForm(Form):
```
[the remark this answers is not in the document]
kvig: This is now a TaskParameterForm

functionary/functionary/urls.py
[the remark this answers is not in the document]
kvig: I was able to override the LOGIN_URL and LOGOUT_URL and set them to the ui application.

functionary/ui/views/functions.py
kvig: The GET being checked for here doesn't really jive with how the form is used so I added a require_POST decorator and removed the request type check.

functionary/ui/urls.py
```diff
@@ -1,41 +1,63 @@
...
+from django.contrib.auth.decorators import login_required
+from django.urls import include, path
 
 from .views import environments, functions, home, packages, tasks, teams
 
 app_name = "ui"
 
 urlpatterns = [
-    path("", home.home, name="home"),
+    path("", login_required(home.home), name="home"),
```
scott-taubman, line 11: This obviously works, but it's slightly confusing to see something clearly labeled as a decorator used this way.  For class based views, it looks like the thing you want is [`LoginRequiredMixin`](https://docs.djangoproject.com/en/4.1/topics/auth/default/#the-loginrequired-mixin), unless there was a specific reason you went this way with it.

functionary/ui/views/environments.py
```diff
@@ -1,28 +1,42 @@
...
         """Sorts based on team name, then env name."""
-
-        return super().get_queryset().order_by("team__name", "name")
-
-
-class EnvironmentDetailView(DetailView):
+        if self.request.user.is_superuser:
+            return super().get_queryset().order_by("team__name", "name")
+        else:
+            environments = EnvironmentUserRole.objects.filter(
```
scott-taubman, line 19: This is the other spot I noticed missing the inheritance of roles assigned at the Team level.  I think I had a suggestion at some point that maybe having a helper on the User model would make sense, such that this could basically just be:

```python
return self.request.user.environments.all()
```

or some such thing.  Neither part (the Team inheritance or the helper) is mandatory for this PR, but at a minimum add a TODO for us to come back and do that.

functionary/ui/views/teams.py
```diff
@@ -1,14 +1,26 @@
...
 
 class TeamListView(ListView):
     model = Team
 
+    def get_queryset(self):
+        """Sorts based on team name, then env name."""
+        if self.request.user.is_superuser:
+            return super().get_queryset().order_by("name")
+        else:
+            teams = TeamUserRole.objects.filter(user=self.request.user).values(
```
scott-taubman, line 17: I guess this is similar to the `EnvironmentListView` situation.  But also I at some point suggested that the way you've got the environments and teams displaying in the UI currently isn't really what we we're going to want, so before you change anything we should probably talk about if those things just go away for now.

functionary/ui/views/view_base.py
```diff
@@ -0,0 +1,49 @@
...
+    model_field = "environment"
+    environment_through_field = None
+    order_by_fields = ["name"]
+
+    def get_queryset(self):
+        """Filters out object not in the environment then sorts based
+        on the value of order_by_fields."""
+
+        env_id = self.request.session["environment_id"]
+        env = Environment.objects.get(id=env_id) if len(env_id) > 2 else None
```
scott-taubman, line 19: What's the `len(env_id) > 2` check about?  What would the env_id be when it's <= 2?

functionary/ui/views/view_base.py
```diff
@@ -0,0 +1,49 @@
...
+        if env:
+            if self.request.user.is_superuser or self.request.user.has_perm(
```
scott-taubman, line 21: The `has_perm` check would be `True` in the superuser case so  you don't need to check that explicitly.

functionary/ui/urls.py
kvig: That's a better way :D

functionary/ui/views/view_base.py
kvig: If the environment isn't selected or the "Choose" value(-1) is selected in the environment selector, the env_id won't be a valid uuid. I was taking the easy way out here.

functionary/ui/views/view_base.py
kvig: done

functionary/ui/views/environments.py
kvig: fixed

functionary/core/models/user.py
```diff
@@ -94,3 +95,20 @@ def environment_permissions(
             permissions.update(self.team_permissions(environment.team))
 
         return permissions
+
+    def environments(self) -> Set["Environment"]:
```
scott-taubman, line 99: I think it makes sense to make this more general and have it just return the queryset.  The ordering and the fact that it's a set are kind of in service of the way you're currently using this, but at the expense of broader applicability.  Also, let's make it a property method so that the way you interact with it is similar to other model relations.

So this method becomes:

```python
@property
def environments(self) - > QuerySet:
    """docstring"""
    Environment = apps.get_model("core", "Environment")

    return Environment.objects.filter(
                models.Q(team__user_roles__user=self) | models.Q(user_roles__user=self)
            ).distinct()
````

And then later on you'd use it like:

```python
envs = self.request.user.environments.order_by("team__name", "name")

for env in envs:
   ...
```

scott-taubman (approved):

PR #57, opened 2022-09-08
Task Execution

### LINKED ISSUE #56: Publish task message to broker on Task instance creation
When an instance of a Task is created, a message should be published to the message broker so that the runner can pick it up and actually execute the task.

scott-taubman:
Closes #56 

This PR adds message publishing to the RabbitMQ broker when a Task instance is created.  This is done via a `post_create` hook on the Task model.

There are also a lot of incidental changes both directly and indirectly related to that.  Comments will be provided on specific lines in the PR.

## Test Instructions
* The READMEs have been updated to include instructions on how to configure the RABBITMQ username and password, as well as how to start the worker and runner processes.  Try following those instructions to ensure that they are accurate.
* Once everything is running, create a new Task.
* You should see the runner listener pick up the message and had it over to the runner celery worker, which will execute the tasking and publish a message with the results.  This should be visible in the logs, but also if you check the "tasking.results" queue in RabbitMQ, there should be a message there.

### DIFF
--- a/functionary/builder/utils.py
+++ b/functionary/builder/utils.py
@@ -19,8 +19,6 @@ from .celery import app
 from .exceptions import InvalidPackage
 from .models import Build, BuildResource
 
-_docker_client = docker.from_env()
-
 logger = get_task_logger(__name__)
 logger.setLevel(getattr(logging, settings.LOG_LEVEL))
 
@@ -102,6 +100,8 @@ def build_package(build_id: UUID):
     Args:
         build_id: ID of the build being executed
     """
+    docker_client = docker.from_env()
+
     # TODO: Catch exceptions and record failures. Also, what happens to the image we
     #       pushed? Should it be deleted?
     logger.info(f"Starting build {build_id}")
@@ -126,17 +126,17 @@ def build_package(build_id: UUID):
     _extract_package_contents(package_contents, workdir)
     _load_dockerfile_template(dockerfile, workdir)
 
-    image, build_log = _docker_client.images.build(
+    image, build_log = docker_client.images.build(
         path=workdir,
         pull=True,
         forcerm=True,
         tag=full_image_name,
     )
 
-    _docker_client.images.push(full_image_name)
+    docker_client.images.push(full_image_name)
 
     logger.debug(f"Cleaning up remnants of build {build_id}")
-    _docker_client.images.remove(image.id)
+    docker_client.images.remove(image.id)
     shutil.rmtree(workdir)
 
     with transaction.atomic():
--- a/functionary/core/utils/messaging.py
+++ b/functionary/core/utils/messaging.py
@@ -1,0 +1,139 @@
+import json
+import logging
+import ssl
+from typing import Tuple
+
+import pika
+from django.conf import settings
+from pika.exceptions import UnroutableError
+from pika.exchange_type import ExchangeType
+
+logger = logging.getLogger(__name__)
+
+PUBLIC_EXCHANGE = "runners.public"
+PUBLIC_QUEUE = "public"
+TASK_RESULTS_QUEUE = "tasking.results"
+
+
+def build_connection(ca=None, cert=None, key=None, open_callback=None):
+    """Creates a connection to RabbitMQ.
+
+    This will use the RABBITMQ_[USER,PASSWORD,HOST,PORT] environment
+    variables to open a connection to RabbitMQ. Optionally pass in
+    certificate information and/or a callback function.
+
+    Args:
+      ca: The path to the CA file for the SSL context
+      cert: The path to the user certifcate
+      key: The path to the keyfile for the given certificate
+      open_callback: If populated, will return a select connection
+        with this as the open callback.
+
+    Returns:
+      A pika.SelectConnection if open_callback is populated, otherwise
+      a pika.BlockingConnection.
+    """
+    connection_params = {
+        "host": settings.RABBITMQ_HOST,
+        "port": settings.RABBITMQ_PORT,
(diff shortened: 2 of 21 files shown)


functionary/builder/utils.py
```diff
@@ -102,6 +100,8 @@ def build_package(build_id: UUID):
     Args:
         build_id: ID of the build being executed
     """
+    docker_client = docker.from_env()
```
scott-taubman, line 103: I moved this because it's previous placement meant that docker was required even for the non-builder processes.

functionary/core/utils/messaging.py
```diff
@@ -0,0 +1,139 @@
...
+from pika.exchange_type import ExchangeType
+
+logger = logging.getLogger(__name__)
+
+PUBLIC_EXCHANGE = "runners.public"
+PUBLIC_QUEUE = "public"
+TASK_RESULTS_QUEUE = "tasking.results"
+
+
+def build_connection(ca=None, cert=None, key=None, open_callback=None):
```
scott-taubman, line 18: There's a nearly identical version of this in the runner code.  I'm going to live with that for now and maybe at some point figure out how to consolidate these things.

runner/runner/celery.py
```diff
@@ -1,8 +1,13 @@
...
+
 from celery import Celery
 
+REDIS_HOST = os.getenv("REDIS_HOST", "localhost")
+REDIS_PORT = os.getenv("REDIS_PORT", "6379")
+
 app = Celery(
     "runner",
-    broker="redis://redis",
+    broker=f"redis://{REDIS_HOST}:{REDIS_PORT}",
```
scott-taubman, line 10: As mentioned elsewhere, this may change to use RABBITMQ, but this branch already does a lot so that would be a separate effort.

runner/runner/handlers.py
```diff
@@ -33,15 +31,13 @@ def __init__(self):
...
+    self._docker.images.pull(package)
 
-    logger.debug(f"Pulled {fqi}")
-    return task
+    logger.debug(f"Pulled {package}")
 
 
 @app.task(base=PackageClient, bind=True)
-def run_task(self, task):
+def run_task(self, _=None, task=None):
```
scott-taubman, line 40: I sort of traded one annoyance (pull_image returning the task only for the benefit of the celery chaining) for another here (having an unused `_` parameter so that the celery chaining ends up with the arguments properly aligned), but I felt like this annoys me ever so slightly less.

runner/runner/listener.py
```diff
@@ -21,51 +22,43 @@ def start_listening():
...
+    logger.info("Connected")
     connection.channel(on_open_callback=_on_channel_open)
 
 
 def _on_channel_open(new_channel):
     """Called when our channel has opened"""
-    logger.debug("channel opened...")
-    global channel
-    channel = new_channel
-    channel.queue_declare(
```
scott-taubman, line 35: I removed the queue declaration stuff here because the runner shouldn't be deciding and creating that stuff.  It should just be able to assume that the queue it's going to listen on has already been created for it.  For now, that's the public runner queue.  Eventually, that'll be a queue established by the main app during runner registration.

runner/runner/handlers.py
```diff
@@ -87,5 +82,5 @@ def _parse_container_logs(logs):
     },
     autoretry_for=(Exception,),
 )
-def publish_result(self, result, topic):
-    send_message(topic, "TASK_RESULT", result)
+def publish_result(self, result):
+    send_message("tasking.results", "TASK_RESULT", result)
```
kvig, line 86: Is this hard code going to be permanent? or do we need a TODO here for later?

kvig (changes requested):
Ran it and it all worked. A couple comments on it.

My biggest concern is that I have 5 terminals open now to run everything which is starting to get unweildy. I'm not sure if there's any way to ease this, maybe having something to run the celery and python listener at the same time?

scott-taubman:
> Ran it and it all worked. A couple comments on it.
> 
> My biggest concern is that I have 5 terminals open now to run everything which is starting to get unweildy. I'm not sure if there's any way to ease this, maybe having something to run the celery and python listener at the same time?

Yeah, I agree, both from a developer perspective and the eventually deployment scenario.

For development, I think I should work on getting a fully fleshed out docker-compose.yaml together somewhat soon-ish so that you have an easy way to start all the things, with most of the configuration taken care of for you.

As for the listener and the workers, I agree that from an end-user perspective of the admin who is responsible for deploying all of these things, the runner is a single thing.  It would be pretty obnoxious to make someone deploy the listener and the worker components separately.  I definitely think the two should be managed by a shared process and eventually would be deployed as a single container.

runner/runner/handlers.py
scott-taubman: No, this shouldn't be hardcoded.  This will probably be part of the info that the runner gets back during the registration process.  I'll add a TODO along those lines.

scott-taubman: TODO Added.

runner/README.md
```diff
@@ -3,3 +3,17 @@
...
+```shell
+# required env variables
+export RABBITMQ_USER=someuser
+export RABBITMQ_PASSWORD=greatpassword
+
+# start the listener
+LOG_LEVEL=INFO python ./main.py
+
+# start the worker
+celery -A runner worker --loglevel=INFO
```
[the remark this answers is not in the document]
scott-taubman: I reworked the README slightly to hopefully make it clearer.

kvig (approved):

PR #58, opened 2022-09-08
Issue 55: misc package creation updates

### LINKED ISSUE #55: Misc Package Creation Updates
Some miscellaneous outstanding updates that are needed to the "package create" command in the CLI:

* The language should be a required argument.  Right now if you don't specify a language, it defaults to python.  This gives python a different status than other languages, which is not what we're going for.
* The yaml file that gets created should be named `package.yaml`, rather than `<package_name>.yaml`
* The `package.yaml` format needs to be updated to reflect the current schema.  We should fill out the package name and have examples / descriptors for the remaining stuff.
* It might be nice to have "next steps" output after creating the package.  So something like:
  ```
  $ functionary package create -l python somethinguseful

  Package something useful created at <path to somethinguseful>

  Next steps:
  * Write your functions in the generated functions.py
  * Update the package.yaml with your package and function information
  * When ready, publish the package to your environment by running:
          functionary package publish <path to somethinguseful>
  ```
  Make good use of color and other formatting to dress that up as you see fit.

catattack05:
Closes #55 

It's not quite there, but it's at the point where it would be good to have some feedback. Specifically, the way I've set things up now is having a template.yaml file to copy over from and then replacing what we need to when we run the create command. The thing I don't currently like is the way the language and name get copied in, so advice on how to do it better is totally welcome


cli/functionary/package.py
```diff
@@ -17,15 +17,16 @@ def create_languages() -> list[str]:
...
-        "name": name,
-        "version": "1.0",
-        "x-language": language,
-    }
-
-    path = pathlib.Path(output_dir).resolve() / name / f"{name}.yaml"
-    with path.open(mode="w"):
-        path.write_text(yaml.dump(metadata))
+    path = pathlib.Path(output_dir).resolve() / name / "package.yaml"
+    path2 = pathlib.Path(
```
scott-taubman, line 21: As written, this will fail if you run the command anywhere outside of a very specific directory.  This should be able to be run from anywhere.  Look at how the `create` command finds the language templates.  You should be able to do the same thing.

On the topic of the language templates, that same folder is probably a more appropriate place for the package.yaml template.  So instead of having a `template.yaml` at the root, I would instead move it to `templates/package.yaml`.

cli/functionary/package.py
```diff
@@ -17,15 +17,16 @@ def create_languages() -> list[str]:
...
-    metadata = {
-        "name": name,
-        "version": "1.0",
-        "x-language": language,
-    }
-
-    path = pathlib.Path(output_dir).resolve() / name / f"{name}.yaml"
-    with path.open(mode="w"):
-        path.write_text(yaml.dump(metadata))
+    path = pathlib.Path(output_dir).resolve() / name / "package.yaml"
```
scott-taubman, line 20: Let's use descriptive names for the paths.  `package_path` and `template_path` rather than `path` and `path2`.

cli/functionary/package.py
```diff
@@ -60,6 +61,13 @@ def create_cmd(ctx, language, name, output_directory):
 
     shutil.copytree(str(basepath), str(dir), dirs_exist_ok=True)
     generateYaml(output_directory, name, language)
+    click.echo(
+        "Next steps: \n"
```
scott-taubman, line 65: The formatting here is sufficiently bland that I completely missed the fact that this was displayed when I originally tested it out.  It just looks like a wall of text that my brain ignores by default.  Try to format it / punch it up in such a way that people will actually take notice of the text there and be less inclined to ignore it.

Also, the first line of the output (which was already there before your changes) is:

> Generating python function named mypackage

This doesn't feel like it tells us anything.  It's basically "Doing the thing you just asked me to do".  It probably makes sense to have the message come after the package is created and be a confirmation that the requested operation was successful.

cli/functionary/template.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+#      # (required) Parameters that the function takes
+#      parameters:
+#          # (required) should match the name of the parameter in the actual function signature
+#        - name: string
+#          # (optional) defaults to name; determines how the parameter is displayed in the UI
+#          display_name: string
+#          # (optional) detailed description
+#          description: string
+#          # (required) data type of the parameter     <-- list of valid types TBD, so leave it at this for now
```
scott-taubman, line 30: The `<--` comments were my comments and not intended to make it into the final template :smile: .

scott-taubman (changes requested):

cli/functionary/package.py
```diff
@@ -17,15 +17,16 @@ def create_languages() -> list[str]:
...
 
 def generateYaml(output_dir: str, name: str, language: str):
-    metadata = {
-        "name": name,
-        "version": "1.0",
-        "x-language": language,
-    }
+    package_path = pathlib.Path(output_dir).resolve() / name / "package.yaml"
+    template_path = (
+        pathlib.Path(__file__).parent.resolve() / "templates" / "template.yaml"
```
scott-taubman, line 22: Let's name the actual file `package.yaml`.  The fact that it's in the `templates` directory now makes it apparent that it's a template.  I suggested that in my last comment, but it probably got last in the wall of text that most of my comments end up being.

cli/functionary/package.py
```diff
@@ -61,6 +60,22 @@ def create_cmd(ctx, language, name, output_directory):
     shutil.copytree(str(basepath), str(dir), dirs_exist_ok=True)
     generateYaml(output_directory, name, language)
 
+    click.echo()
+    click.echo(f"Package creation for {name} successful!\n")
+    text = Text()
+    console = Console()
+    text.append("Next Steps: \n", style="b u blue")
```
scott-taubman, line 67: I think this looks much better now.  A few super minor suggestions.

* "Next Steps: " I think you can drop the colon.  I'm 50/50 on if the underline is necessary.  The bold and the color choice make it stand out enough to me, but I'll let you decide if the underline stays or goes.
* Drop the trailing space before the newline on each of these. (e.g. `"Next Steps\n"` rather than `"Next Steps \n"`)
* Let's add an additional newline before the publish command line.  I think that helps make it more readable.

scott-taubman (changes requested):

scott-taubman (approved):

PR #60, opened 2022-09-14
Issue 59: Some minor package create cleanup

### LINKED ISSUE #59: Clean up CLI package create
Two issues noticed during CLI demo today:
-Package.yaml shows up on the list of possible languages to use for -l
-We probably don't want to let people create a file name with the '/' character; if they do, display a message saying something along the lines of "This looks like a path. Did you mean to do package create -o <path> <filename>?'

catattack05:
Closes #59 

Just some small edits to have the language option skip over printing the yaml file and also to make any input name with a '/' stop to give the user better directions.

### DIFF
--- a/cli/functionary/package.py
+++ b/cli/functionary/package.py
@@ -13,7 +13,11 @@ from .utils import format_results
 
 def create_languages() -> list[str]:
     spec = pathlib.Path(__file__).parent.resolve() / "templates"
-    return [str(loc.name) for loc in spec.glob("*")]
+    languages = []
+    for loc in spec.glob("*"):
+        if str(loc.name) != "package.yaml":
+            languages.append(str(loc.name))
+    return languages
 
 
 def generateYaml(output_dir: str, name: str, language: str):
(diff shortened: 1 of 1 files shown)


cli/functionary/package.py
```diff
@@ -51,6 +55,13 @@ def create_cmd(ctx, language, name, output_directory):
 
     Create an example function in the specified language.
     """
+    if "/" in name:
+        ex_path = name.rsplit("/", 1)[0]
+        ex_name = name.rsplit("/", 1)[1]
+        raise click.ClickException(
+            "Your package name looks like a path. Try using this command instead:"
+            f"\n       functionary package create -l {language} -o {ex_path}/ {ex_name}"
```
scott-taubman, line 63: Let's drop the trailing slash in `{ex_path}/`.  With it there, if you copy and paste the command as it gets output, you'll end up with a "//" in the path that gets output at the end.

scott-taubman (changes requested):

scott-taubman (approved):

PR #68, opened 2022-09-19
Issue 62: Add functions dynamically to yaml

### LINKED ISSUE #62: Dynamically generate functions portion of package.yaml
Some portions of the function definitions that must be defined in the package.yaml could be generated based on the function signatures in the functions.py.  There should be a command available in the CLI that updates the package.yaml for you with whatever it can glean from those functions.

For example.  Let's say my functions.py looks like this:

```python
def add(x: float, y: float = 5.0):
    """Adds two numbers"""
    pass

def repeat(word: str, count: int = 2, loud: bool = False):
    """Repeat a word, possibly often, possibly loudly"""
    pass
```

The goal is to be able to run a command like:

```shell
functionary package genschema /path/my/package
```

And update the functions portion of our package.yaml to look like:

```yaml
  functions:
    - name: add
      description: Adds two numbers
      parameters:
        - name: x
          type: float
          required: true
        - name: y
          type: float
          default: 5.0
          required: false
    - name: repeat
      description: Repeat a word, possibly often, possibly loudly
      parameters:
        - name: word
          type: string
          required: true
        - name: count
          type: int
          default: 2
          required: false
        - name: loud
          type: bool
          default: false
          required: false
```

catattack05:
Closes #62 

First draft attempt to populate the yaml automatically. I'm using the inbuilt python libraries inspect and importlib.util to import the module the user passes in and parse the module/function objects that inspect returns. At the end, I'm replacing a placeholder in the yaml with the json.dumps (though it's a little awkwardly formatted when it shows up on the yaml, so that's an area I could use some tips on)

### DIFF
--- a/cli/functionary/parser/python.py
+++ b/cli/functionary/parser/python.py
@@ -15,28 +15,6 @@ type_map = {
 }
 
 
-def _retrieve_function_data(path):
-    """
-    Helper function for py_parse that
-    gets function data from path/function.py
-
-    Args:
-        path: the path to file
-    Returns:
-        function_data: content of file as string
-    Raises:
-        ClickException if file not found or cannot be accessed
-    """
-    try:
-        with open(path + "/functions.py", "r") as func_file:
-            function_data = func_file.read()
-    except FileNotFoundError:
-        raise click.ClickException("Could not find functions.py")
-    except PermissionError:
-        raise click.ClickException("Did not have permission to access functions.py")
-    return function_data
-
-
 def _retrieve_function_nodes(data):
     """
     Helper function for py_parse to retrieve
@@ -94,23 +72,25 @@ def _create_argument_list(node):
             except KeyError:
                 type = None
 
+        # Set as required now, will update later if needed when handling defaults
         arg_dict["required"] = True
         arg_list.append(arg_dict)
 
         if type is None:
             click.echo(
-                f"Type of argument {arg.arg} in function {node.name} not detected or "
-                "is unsupported. Please specify type in package.yaml manually before "
-                "publishing package, as argument type is required.\n"
+                f"The argument type of {arg.arg} in function {node.name} could "
+                "not be parsed, was not specified, or is an unsupported type. "
+                "Please update package.yaml with arg type manually prior to "
+                "publishing package to ensure a successful build.\n"
             )
     return arg_list
 
 
 def _assign_defaults(node, arg_list):
     """
-    Helper function for py_parse. If arg
+    Helper function for py_parse. If arg in list
     has a default, fill out arg's dictionary with
-    default
+    that default
 
     Args:
         node: ast node representing function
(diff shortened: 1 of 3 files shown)


scott-taubman:
I think we're going to want to take another crack at this.  A few very high level comments / things to try:

* Let's take the `populate_python` and `populate_js` functions out of the package.py.  Instead, call some helper that simply receives the language (and probably the path to the function file) as input and returns the dict for the functions as the output.  Basically, we want to abstract away all of the language specific bits, including even which function needs to get called (as future non-python languages may not be a function call at all, but a call to an external process).
* For python, ideally we could get the job done without actually have to load the module.  Loading the module means that any dependencies would need be installed first before we can parse the functions, since those dependency would get imported when the module is loaded.  While that's not a deal-breaker necessarily, I think we should try to avoid it.  Look at using the [`ast`](https://docs.python.org/3/library/ast.html) module.  You can use [`ast.parse()`](https://docs.python.org/3/library/ast.html#ast.parse)  on a string of the code and then find the functions from there.  
* Don't worry about attempting to preserve the comments in the `package.yaml` during this process.  The comments are there initially to help guide someone who is manually updating the file.  If we lose the comments as a tradeoff of automatic the function generation, that's acceptable.  I say this because there's a part where you are loading the yaml as a string and parsing it.  Rather than doing that, just load the yaml to a dict and work with it programatically.
* We haven't defined what the mapping is going to be yet, but don't assume that the python types are what you're going to want to write into the package.yaml as the types.  Instead, create a type map and do a lookup.  If you want to have the map just be the python types for now (str --> str, int --> int, etc.) that's fine.  But we'll shortly be deciding what that mapping looks like and we'll want to be able to just go to the map and update it, rather than having to dive into the bowels of parsing code.

scott-taubman:
Some general notes:

* The parse function that is in the package.py can be pulled out into it's own file.
* The language specific bits could be their own files as well, since the handling across different languages is likely be very different.
* The yaml file can be loaded using yaml.load now since you don't need to worry about trying to preserve the comments and all that.
* In the parse function, use the language to find the function that handles the parsing for that language and just call that function, passing all of the relevant data along.  If no function exists to handle that language, output and error along the lines of "Schema generation for {langauge} is not currently supported".
* You should be able to use slicing to consolidate the two blocks of the defaults handling, rather than having different paths for when the default and arglist count are and are not the same.
* Always have the arglist be a list, even if it's empty.
* The types for the type map have now been roughly established, so you can use real values there.  See this map [here](https://github.com/beer-garden/functionary/blob/main/functionary/builder/api/v1/serializers/package_definition.py#L5).  Most of the things that map are obvious (int -> integer, str -> string, etc.). The only one that's not a direct map would be dict -> json.

scott-taubman:
* Squash and rebase onto main just to get all of the latest stuff that's been merged since this was opened.
* The parsing code can go in its own folder  to make the relationship between things clearer.
  * The code that's in parse.py can go in parser/__init__.py.
  * py_parse.py can become parser/python.py
* The type mappings should match the first item in the tuple for each of the PARAMETER_TYPES defined in functionary/builder/api/v1/serializers/package_definition.py.
* Try to break the code that's currently in py_parse.py up into smaller functions to help with readability.  It's difficult to follow as it currently is.
* It should be possible to do your for loops in such a way that you don't need the separate `i` tracker for the index you're working at.
* I don't think you need the check of whether the length of defaults and arg_list are the same.  If they are equal, the `def_start` would just be 0 and everything would get set the way you're intending anyway.
* It should be okay to just treat unknown parameter types the same as if the type is not set.  Just ignore it and don't populate the type in the yaml.
* Some of the exception messages in the package.py genschema have typos / weird wording.

cli/functionary/parser/python.py
```diff
@@ -0,0 +1,191 @@
...
+        now with defaults filled out as needed
+    """
+    defaults = node.args.defaults
+
+    default_start = len(arg_list) - len(defaults)
+    has_default = arg_list[default_start:]
+    no_default = arg_list[:default_start]
+
+    for arg, default in zip(has_default, defaults):
+        if arg["type"] == type_map["date"]:
```
kvig, line 129: If `arg["type"]` doesn't exist just use default.value, otherwise this line explodes.

kvig (changes requested):
Just a couple more comments and it should be ready.

scott-taubman (approved):

PR #69, opened 2022-09-19
#67 - process, store, and display tasking results

### LINKED ISSUE #67: Process and store task results
This is the general effort of processing the results messages that come back from the runner and storing the task results in the database.

scott-taubman:
Closes #67 

This PR adds a **very basic** implementation of function execution results handling.  It simply does the following:

* Adds two new models: `TaskResult` and `TaskLog`.
* Adds a listener to consume the messages produced by the runner.
* Hands `TASK_RESULT` messages off to a new `record_task_result` celery task.
* That task writes the function results to `TaskResult` and the non-result output to `TaskLog`.

To be clear, this is not an especially scalable solution, but it provides a basic means of showing results for now.

## Test Instructions
The docker-compose file and VSCode config has been updated to include the new listener process.  Just run the "docker-compose up" task and everything that's needed should start.

* Task something via the UI.
* It should redirect you to the task result page upon submission.
* Sit tight for a few seconds and the results should appear.
* You can also use the swagger docs to query for the results or logs via the API.


functionary/core/api/v1/serializers/team.py
```diff
@@ -21,6 +22,7 @@ class Meta:
         model = Team
         fields = ["id", "name", "environments"]
 
+    @extend_schema_field(TeamEnvironmentSerializer(many=True))
```
scott-taubman, line 25: Unrelated to the rest of the PR.  This is just fixing an error I found in the swagger docs.

functionary/core/utils/listener.py
```diff
@@ -0,0 +1,55 @@
+import json
```
scott-taubman, line 1: This whole file is again duplicative of the runner listener code.  Both this and the runner code will be refactored at some point so that they share a singular listener / worker code-base.

functionary/ui/templates/core/task_detail.html
```diff
@@ -1,39 +1,55 @@
...
+                <label class="label" for="params">
+                    <i class="fa fa-list"></i>&nbsp;Parameters:
+                </label>
+                <ul id="params">
+                    <li class="block ml-4">
+                        <p class="is-font-monospace">{{ task.parameters }}</p>
+                    </li>
+                </ul>
+            </div>
+            <div class="field">
```
scott-taubman, line 45: This file got completely reformatted by djlint, which is why the changeset is somewhat large.  This div is really the only new bit.

I went with the dumbest possible approach here, which just basically replicates the way we're using unicorn for the tasking status.  The problem with this is that there are now two independent polling events to retrieve information that is ultimately derived from the same source (the task).  We should become more familiar with unicorn and figure out how to consolidate these into a single event, but for now I just took the easy way out.

functionary/ui/components/task_title.py
```diff
@@ -1,16 +1,16 @@
...
+        super().__init__(*args, **kwargs)
+        self.status = self.task.status
 
     def get_status(self):
-        # Get the ID from the end of the referer URL, alternatively
-        # you can parse the body each time and access data.task.pk
-        task_id = self.request.headers["Referer"].split("/")[-1]
-        self.status = Task.objects.values("status").get(id=task_id)["status"]
+        self.task.refresh_from_db()
+        self.status = self.task.status
```
kvig, line 13: Good catch, I'm not sure how I missed that, I was in the debugger too :D

functionary/ui/templates/core/task_detail.html
kvig: I would have updated the endpoint to return the whole task information since there's more than just the status to be updated. This works too and ends quickly enough when the result comes back.

kvig (approved):
Once comment, free to merge.

functionary/ui/templates/core/task_detail.html
scott-taubman: Yeah, that probably makes sense.  I worry about how that scales on pages with a lot of data, but you only need to dynamically refresh a couple of independent pieces of it.  In this case, that's not a concern though so I just consolidated the `task_title` and `task_result` components into a single `task_detail` so that there's only a single polling event now.

kvig (approved):

PR #71, opened 2022-09-20
Add additional parameter types.

kvig:
This PR adds support for the following datatypes:
- string (assumed short input)
- text (longer form input)
- integer
- float
- boolean
- date
- datetime
- json

The json field is parsed and validated that it is json. This allows for object or lists of JSON supported types to be passed to the function.

All the fields are validated to be of the correct datatype and the page will display error messages below the failing fields.

This was tested by making a function with all the possible parameter types and printing the value received by the function in the docker runner.

### DIFF
--- a/functionary/ui/forms/forms.py
+++ b/functionary/ui/forms/forms.py
@@ -36,10 +36,18 @@ _field_mapping = {
     "json": (JSONField, Textarea),
 }
 
-_transform_intitial_mapping = {"json": json.loads}
+_transform_initial_mapping = {"json": json.loads}
 
 
-def get_param_type(param_dict):
+def _get_param_type(param_dict):
+    """Finds the type of the parameter from the definition in the schema.
+
+    Pydantic maps to python types correctly, but we want to use the actual
+    schema type for input purposes. If there is a format in the param_dict,
+    the input field has a definition of what type it is. If there is an anyOf,
+    inspect it and try find out what type it should be. Otherwise, return the
+    value of 'type' in param_dict.
+    """
     keys = param_dict.keys()
 
     # pydantic hides the date type in the format field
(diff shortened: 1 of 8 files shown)


functionary/ui/forms/forms.py
```diff
@@ -31,20 +66,28 @@ def __init__(self, function, data=None):
...
+            field_class, widget = _field_mapping.get(param_type, (None, None))
 
             if not field_class:
-                raise ValueError(f"Unknown field type for {param}: {value['type']}")
+                raise ValueError(f"Unknown field type for {param}: {param_type}")
+
+            if widget:
+                kwargs["widget"] = widget
+            if param_type in _transform_intitial_mapping:
+                kwargs["initial"] = _transform_intitial_mapping[param_type](initial)
```
scott-taubman, line 88: This call throws an exception if initial is `None`, so add that as part of your check before reaching this line.

Also, this might just be a me thing, but I keep reading `_transform_initial_mapping` as an action, which then makes my brain want to think that that's a function.  Which, maybe it could be?  What if instead you did:

```python
if widget:
    kwargs["widget"] = widget

kwargs["initial"] = _prepare_initial_value(param_type, initial)
```

And then the `_prepare_initial_value` call has all of the conditional logic, etc.  I know it's all kind of the same in the end.  Making it a function also has the added benefit of giving you a natural place for a simple docstring to explain what it's all about :-).

functionary/ui/forms/forms.py
```diff
@@ -1,28 +1,63 @@
...
+        DateTimeField,
+        HTMLDateTimeInput,
+    ),
+    "json": (JSONField, Textarea),
 }
 
+_transform_intitial_mapping = {"json": json.loads}
+
+
+def get_param_type(param_dict):
```
scott-taubman, line 42: Let's add a simple one line docstring here (and places like it going forward).  Also, for the sake of consistency, if this is not intended to be used outside of this file, let's prepend the name with an underscore.

scott-taubman (commented):
Mostly looks really good.  Commented about a couple of errors I ran into and some nitpicks.

functionary/ui/forms/forms.py
```diff
@@ -31,20 +66,28 @@ def __init__(self, function, data=None):
         super().__init__(data)
 
         for param, value in function.schema["properties"].items():
-            req: bool = not value["default"]
+            req = not value["default"]
```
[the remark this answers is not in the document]
kvig: Fixed

functionary/ui/forms/forms.py
kvig: Fixed

functionary/ui/forms/forms.py
kvig: Fixed

PR #74, opened 2022-09-21
Issue 73: add summary field for package and function

kvig:
This PR adds optional fields for summary for both Packages and Functions. 

The calculator example is updated with the new summary field, the template has a stub for the summary field also. The database has the new optional summary field and the UI has been updated to show the summary on the Package/Function lists as well as the Environment details. Package/Function description is shown on their details page.

The CLI create command was updated to prompt for language if that required option was missing from the command line.

### DIFF
--- a/cli/functionary/package.py
+++ b/cli/functionary/package.py
@@ -41,6 +41,7 @@ def package_cmd(ctx):
     "-l",
     type=click.Choice(create_languages(), case_sensitive=False),
     required=True,
+    prompt="Select the language",
 )
 @click.option("--output-directory", "-o", type=click.Path(exists=True), default=".")
 @click.argument("name", type=str)
--- a/cli/functionary/package.py
+++ b/cli/functionary/package.py
@@ -140,8 +140,12 @@ def list(ctx):
         function_dict = {}
         function_dict["Function"] = function["name"]
         function_dict["Display Name"] = function["display_name"]
-        function_dict["Summary"] = function["summary"]
-        function_dict["Description"] = function["description"]
+
+        # Use the summary if available to keep the table tidy, otherwise
+        # use the description
+        if not (description := function.get("summary", None)):
+            description = function["description"]
+        function_dict["Description"] = description if description else ""
 
         if package_id in functions_lookup:
             functions_lookup[package_id].append(function_dict)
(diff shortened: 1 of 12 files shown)


scott-taubman:
Alright, I did vaguely say "short", but 50 feels **really** short.  We allow the name to be bigger than that :-P.  The goal is to allow a short, one-line blurb to display in places like the list views, the CLI output, etc.  What's a good size limit for that?  At least 128 and no more than 256 would be my totally arbitrary guess.  Basically short enough to present nicely in the appropriate places, long enough to still be useful.

functionary/ui/templates/core/function_detail.html
```diff
@@ -1,55 +1,65 @@
...
-    <div class="p-3 accordions">
-        <div class="accordion is-info {% if not form %} is-active {% endif %}">
-            <div class="accordion-header toggle">
-                <h3 class="is-4">Schema</h3>
-                <button class="button toggle"></button>
-            </div>
-            <div class="accordion-body">
-                <div class="accordion-content">
-                    <pre>{{ function.schema | pretty_json }}</pre>
+        <div class="field">
```
scott-taubman, line 26: I think having the summary and description both display would be nice.  Mostly because they're both optional, so if someone just has a summary and not a longer form description, as is you don't get anything when you're on the detail page.  Seeing the summary still feels useful there.

Same goes for the package detail view.  In that case, the team name is kind of redundant since you're always operating in the context of a single environment, so maybe take that space and put the summary there instead?

kvig:
I went off the git title limit, upped the summary to 128. Added in the summary if available as well as fixed the lack of using the display_name for Packages and Functions.

cli/functionary/package.py
```diff
@@ -149,9 +151,11 @@ def list(ctx):
     for package in packages:
         name = package["name"]
         id = package["id"]
+        summary = package["summary"]
```
scott-taubman, line 154: For the CLI, let's just show the summary and not the description anymore in the list views.   Basically it'll be similar to the UI then and we don't have to worry about some super long full description causing the list output to get out of hand.

scott-taubman: Also, if there is no summary it ends up displaying "None" right now.  Maybe we can fix that.

PR #80, opened 2022-09-26
#79 - Added fixtures and init script for doing initial provisioning of…

### LINKED ISSUE #79: Add Fixtures for Bootstrapping Functionary Users, Teams, and Environments
When deploying Functionary for the first time, the UI expects some data to already be present, such as a user, team, and environment. By adding fixtures, we can bootstrap an admin user, dev team/environment, so that the user does not need to exec into the Functionary-django container,  `createsuperuser`, and manually create the teams/environments themselves.

xeonneon:
Closes #79 

This PR adds an init script to the `functionary-django` image, such that the user can execute the script to load a bootstrap fixture into the database. This allows for new users to provision a new admin user, team, and environment, without having to create a new admin user themselves to handcraft the users/team/environments from the admin UI.

## Testing
- If using the vscode workspace, run the `functionary: docker compose up` task 
- If not using the vscode workspace, reference the task commands to grab the docker-compose command
- Once everything comes up, run `docker exec functionary-django ./init.sh` from your terminal
- You should see all the migrations get applied to the database, and several objects get installed from the fixture
- You should now be able to log into functionary using the default admin login credentials:
  - Username: `admin`
  - Password: `admin`

## Implementation Notes
- It may make sense to change the Dockerfile's entrypoint to just run an entrypoint script, rather than passing environment vars that have the commands we want to execute
- I am unsure whether or not the fixtures should be automatically run every functionary deployment. I opted for it to be a manually triggered event. If we want it to be automatic, then we should also switch over to the entrypoint script method for the Dockerfiles


xeonneon:
Added VSCode task that runs the `docker exec` command from above for applying migrations and fixtures

PR #84, opened 2022-09-27
Issue 64: Check if directory is empty and avoid overwrite

### LINKED ISSUE #64: Package Create overwrites/appends to existing files
I mixed up my commands and ran `functionary package create -l python kitchen_sink` when I already had a kitchen_sink directory. The create command overwrote my `functions.py` and appended a new yaml to the end of `package.yaml`. The command should be consistent in what it does, and either prompt to overwrite or fail and force the command to run again with a `-f` flag to overwrite what's there.

catattack05:
Closes #64 

The behavior Scott and I discussed was having package create simply not do anything if the directory the user wants to access isn't empty. So that's what I've added here

### DIFF
--- a/cli/functionary/package.py
+++ b/cli/functionary/package.py
@@ -65,6 +65,7 @@ def create_cmd(ctx, language, name, output_directory):
     elif os.listdir(dir):
         raise click.ClickException(
             f"Create command failed: {output_directory + '/' + name} is not empty."
+            "Destination must be a new or empty directory."
         )
 
     basepath = pathlib.Path(__file__).parent.resolve() / "templates" / language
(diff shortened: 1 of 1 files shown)


cli/functionary/package.py
```diff
@@ -61,6 +62,11 @@ def create_cmd(ctx, language, name, output_directory):
     if not dir.exists():
         dir.mkdir()
 
+    elif os.listdir(dir):
+        raise click.ClickException(
+            f"Create command failed: {output_directory + '/' + name} is not empty."
```
scott-taubman, line 67: Let's have this say:

```suggestion
            f"{output_directory + '/' + name} is not empty. Destination must be a new or empty directory."
```

scott-taubman (commented):
Made a suggestion for slightly different wording on the message, but otherwise looks good.

PR #89, opened 2022-10-04
Establish basic unit test structure and some examples

### LINKED ISSUE #88: Bootstrap unit tests
We need to start building up a suite of tests.  To get started, we should do the following:

* Establish a basic structure / layout of tests.
* Ensure the django unit tests are easily runnable through vscode
* Configure a basic github action for running the testsgit

scott-taubman:
Closes #88 

This PR establishes some basic structure for unit tests, with a smattering of examples.

* For the django project, tests are housed in the individual app directories and mirror the structure of the app itself
* cli has a top level tests folder. It contains a single test, and a conftest.py with a couple of fixtures that will likely be generally useful.
* No tests have been created for the runner code yet.
* A github action has been added to run the tests on PRs.
* The vscode settings have been updated to enable pytest.  The tests should be auto-discovered, but if they aren't just do the basic test config and tell it you are using pytest.

Other notes:
* I slightly re-organized the settings files, converting base.py into its own module.  This was so that the environment configs (debug, test, prod) are more clearly distinguishable from the base config files.

## Testing Instructions
* Be sure to pip install the new dependencies, specifically the requirements files in the cli and functionary folders.
* From the top level directory you can run:
  * pytest cli
  * pytest functionary
* You should also make sure you can use the testing component in vscode to run the tests.


functionary/core/tests/api/v1/views/test_task.py
```diff
@@ -0,0 +1,68 @@
...
+        "parameters": {"prop1": 5},
+    }
+    response = admin_client.post(
+        url, data=task_input, content_type="application/json", **request_headers
+    )
+    task_id = response.data.get("id")
+
+    assert response.status_code == 201
+    assert task_id is not None
+    assert Task.objects.filter(id=response.data["id"]).exists()
```
kvig, line 50: Did you want to reuse task_id here?

kvig (approved):
Looks good and works for me. Two small comments, you can fix or merge as desired.

functionary/core/tests/api/v1/views/test_task.py
scott-taubman: That would have made sense.  Done.

cli/tests/conftest.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+
+from functionary.config import config_file
+
+
+@pytest.fixture
+def fakefs(fs):
+    """Wrapper around the pyfakefs fs fixture. This is provided slightly increased
```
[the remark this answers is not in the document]
scott-taubman: I think I was going for "This is **provided for** slightly increased..." Which I have now changed it to.

The distinction being I'm only explaining what the goal was, not making an assertion that I was successful in said goal 😜.

Issue #81, opened 2022-09-26
Remove docker-in-docker dependency for the builder


xeonneon:
Remove the builder's requirement for docker-in-docker, so the host's docker socket is not being mounted to the containers.

Use some other tool for building and running containers that has a Python SDK available.

xeonneon:
## Introduced Changes
- Two new images: `functionary_django`, `functionary_podman_django` 
  - Only the functionary components that need podman should use the `functionary_podman_django` image
- Switched base images from `python:3.10-slim` to `python:3.10-alpine` to save space 
  - slim: `126MB`
  - alpine: `48.7MB`
- Images now run completely rootless, using the `functionary` user with UID `1000`
- Removed python virtual environments from the images, since the containers will already be in an isolated environment
- Added `entrypoint.sh` scripts that the images execute. Uses environment variables to determine which component to start
  - This allows for us to wrap additional bootstrapping logic to any particular component
  - In my case, I use it to create a `podman.sock` when the builder component is run
- Podman runs rootless in the `functionary_podman_django` image
  - Currently, only the builder is using Podman until further testing is done with the interactions between the Docker Python SDK and the `podman.sock`

PR #96, opened 2022-10-06
#81 - Configured and enabled rootless builder and runner

xeonneon:
# Introduction
From a security perspective, it is generally not a great idea to mount the `docker.sock` into containers, since that essentially allows the user's inside the containers (which either need to be `root` in order to interact with the socket, or a member of the same GID as the `docker` group on the host) to run workloads as root on the host. As such, it may make sense to instead use rootless Podman within the containers, such that Functionary can run entirely rootless and not rely on exposing the host's `docker.sock` to the containers. This way, since user's do not need to be root, the unprivileged user's in the containers are mapped to some higher non-privileged UID on the host, which reduces the possibility of a user from breaking out as `root` onto the host.

## Introduced Changes
- Separated the Functionary image into two different images: one with Podman installed, and another without Podman installed
  - Both images run as a non-privileged user `functionary` with UID `1000`
- The two containers that use the Podman-Functionary image are the `build-worker` and the `runner-worker`
  - Both containers will create a `podman.sock` that will be used in place of the `docker.sock`
  - These two containers are granted additional capabilities, which are required for rootless Podman to build and run containers. 
  - They DO NOT use nor require `privileged` mode in order to function.
- Added `entrypoint.sh` scripts to handle the launching of the different services. This allows us to perform additional configurations before launching a service if needed
- Runner launched images now use `v1.0.0` tag instead of `latest`, to prevent unnecessary pulling of the image

## Set Up Instructions
On host: `sudo apt install fuse-overlayfs`
Push the necessary images to the local registry for the deployment:
- Spin up Functionary
- `exec` into the `build-worker` container
- `cd templates/`
- Run `./build.sh -p`
- Select whichever repository you want to pull the image down

## Testing
- Verify user in the container is not root
  - `exec` into one of the containers
  - Run `whoami`
  - Verify result is `functionary` 
- Verify user in container is not running as root on the host
  - `exec` into one of the containers, and run `sleep 100`
  - On the host, run `ps -af | grep sleep`
  - Verify that the user running that process is not `root`
- Verify podman is installed in the `build-worker` and `runner-worker` containers
  - `exec` into both containers
  - Run `podman info`
  - Verify podman spits out a bunch of info about it's installation
- Verify `podman.sock` exists in the `build-worker` and `runner-worker` containers
  - `exec` into both containers
  - `ls -al ~`
  - Verify socket exists
- Verify `podman` can run containers in the `build-worker` and `runner-worker` containers
  - `exec` into both of the containers
  - Run `podman run alpine echo hello world`
  - Podman should pull the latest alpine image, spin up the container, and print out "hello world"
- Verify `build-worker` can build and push images to the registry
  - Use functionary CLI to publish a package to functionary
  - View the logs of the `build-worker`, verify the build starts and succeeds
  - You can `exec` into the `build-worker` container and run `podman image ls` to verify image exists
- Verify `runner-worker` can execute tasks
  - From the Functionary UI, launch a task
  - View logs of the `runner-worker`, verify the runner pulls the necessary image, creates the container, and runs the container
  - From the Functionary UI, verify the task status updates to `Complete`, and the result is expected


scott-taubman:
General notes:

* Switch back to the slim image rather than using alpine
* Let's focus on getting the builder to do the rootless podman thing and allow the runner to continue using the docker sock connection of the host.
* The containers that don't interact with docker in any way can be updated to not run as root.  Let's make the UID a build arg though rather than hardcoding it to 1000.
* The v1.0.0 tag change shouldn't be needed.  Let's just stick with latest.
* The build worker shouldn't need the cap-add stuff or the fuse mount.
* Let's verify the security_opts as well.
* Some of what's happening in the podman.Dockerfile may be specifically in service of spinning up containers, which the build worker doesn't need to do.  We should try to streamline what we can for that context.
* Let's do this for the Dockerfiles:
  * The runner uses the existing debug.Dockerfile (pick a new name for it)
  * The builder uses the new podman.Dockerfile
  * Everything else uses what is currently (in your branch) named functionary.Dockerfile
* You don't need to mount in the template stuff into the containers

scott-taubman:
I'm going to close out this PR.  Some of what this PR was aiming for (non-root users in the containers where possible, a proper entrypoint script) were used as inspiration for changes in #104.  Having the builder and runner operate completely rootless is something that will be revisited later when we look to create cloud native iterations of those components.

PR #99, opened 2022-10-12
Issue 98: Add codes to 400 errors

### LINKED ISSUE #98: Add error codes to API responses
Currently the API only includes the detail message in error responses.  For instance, if you fail to include the required header, you'll get a response like:

```json
{
  "detail": "X-Environment-Id or X-Team-Id header must be set",
}
```

That "detail" message is mostly intended for human consumption.  For something that is programmatically accessing the API, a more parser friendly code should be included so that there is a means of differentiating between the various errors that might share a status code (i.e. the many flavors of 400 errors) without having to resort to parsing the detail message.

What we'd like the response to look like instead is:

```json
{
  "detail": "X-Environment-Id or X-Team-Id header must be set",
  "code": "missing_header"
}
```

catattack05:
Closes #98 

First draft of outfitting the 400 errors with error codes by adding a new exception handler (defined in core/api/exceptions.py for use in settings file functionary/settings/base/rest_framework.py). Current exceptions are InvalidPackage (for package yaml errors, defined in builder/exceptions.py for use in builder/api/v1/views/publish.py ), MissingEnvironmentHeader, InvalidEnvironmentHeader, InvalidTeamIDHeader (all defined in core/api/exceptions.py for use in core/api/mixins.py).

I left BadRequest defined in core/api/exceptions.py, but replaced it with the other exceptions wherever else it is in the code. Left error messages mostly the same

### DIFF
--- a/functionary/core/api/exceptions.py
+++ b/functionary/core/api/exceptions.py
@@ -12,7 +12,8 @@ def custom_exception_handler(exc, context):
         -context: dictionary containing any extra
          context such as the view currently being handled
     Returns:
-        -response: Either a response object or None
+        -The exception response with the additional error code.
+        Unhandled exceptions return None, resulting in a 500 error
     """
     response = exception_handler(exc, context)
     if response is not None:
--- a/functionary/core/api/exceptions.py
+++ b/functionary/core/api/exceptions.py
@@ -8,12 +8,12 @@ def custom_exception_handler(exc, context):
     to response.
 
     Args:
-        -exc: the exception to handle
-        -context: dictionary containing any extra
-         context such as the view currently being handled
+        exc: the exception to handle
+        context: dictionary containing any extra context such as the view currently
+                 being handled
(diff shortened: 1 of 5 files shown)


functionary/core/api/exceptions.py
```diff
@@ -1,4 +1,25 @@
...
+    Exception handler to add code from custom exceptions
+    to response.
+
+    Args:
+        -exc: the exception to handle
+        -context: dictionary containing any extra
+         context such as the view currently being handled
+    Returns:
+        -The exception response with the additional error code.
+        Unhandled exceptions return None, resulting in a 500 error
```
scott-taubman, line 16:
```suggestion
        exc: the exception to handle
        context: dictionary containing any extra context such as the view currently
                 being handled
    Returns:
        The exception response with the additional error code.  Unhandled exceptions
        return None, resulting in a 500 error.
```

PR #100, opened 2022-10-13
#79 - Added additional fixture data

scott-taubman:
This PR builds on the initial work for #79.  It fleshes out the fixture data so that it now contains

* Two separate Teams:
  * FunctionFoundry, containing a dev and prod environment
  * ScriptSanctuary, containing just a default environment
* A bunch of users with various roles for FunctionFoundry:
  * admin: The same superuser account that existed before
  * ff-admin: admin role for the entire team
  * ff-readonly: read only role for the entire team
  * ff-developer: developer role for dev
  * ff-operator: operator role for prod
* The password for all users is the same as their username

The idea is just to have a handful of accounts ready to go with different permissions to easily test or demonstrate access control in various areas.

Additionally, I added a dependent task that starts the django container so that it's now possible to run the "apply migrations and load fixtures" task even if the containers aren't already running.  In that situation the necessary container will be started for you and then the migration task will run.

### DIFF
--- a/docs/VSCODE.md
+++ b/docs/VSCODE.md
@@ -32,7 +32,9 @@ mostly include autoformatting settings, including format on save to ensure that
 code always adheres to those standards. You can see the various settings in the
 `functionary.code-workspace` file.
 
-## Docker Compose
+## Workspace Tasks
+
+### Docker Compose
 
 There are several processes and dependent services that are required to run in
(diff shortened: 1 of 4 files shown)


kvig (approved):
Tried without and then with the migrations step run and the containers operated as expected.

PR #102, opened 2022-10-13
#90 - Initial combination of worker and listener into one entity

### LINKED ISSUE #90: Make the runner listener and worker a single entity
The runner is currently comprised of two distinct processes: a listener and a worker.  It would be preferable if these two processes could be spawned from a single parent process.  That would provide a natural path for consolidating down to a single, self-contained container that is just the "runner", rather than having separate "runner-listener" and "runner-worker" containers.

This is hopefully just a straightforward fork or spawn using [multiprocessing](https://docs.python.org/3/library/multiprocessing.html).  Bonus points for setting the process names nicely.

xeonneon:
Closes #90 

This PR combines the `runner-worker` and `runner-listener` containers into a single container `functionary-runner`. The listener and worker are now `multiprocessing.Process` subclasses, each of which gets spun up from `runner.py`. Since Celery also uses `multiprocessing` for its workers, I configured Celery's logger to print all logging to stdout, so that they are visible when inspecting the container's logs. This avoids needing to do nested multiprocess logging.

## Introduced Changes
- Introduced new `runner.py` which spawns a `Worker` and `Listener` process
- The original functionality for `worker.py` has been turned into a `multiprocessing.Process` subclass
- The original functionality for `listener.py` has been turned into a `multiprocessing.Process` subclass
- Celery now streams its logs to `stdout` so the logs can be viewed when inspecting the container's logs

## Testing
- Spin up Functionary
- Inspect the `functionary-runner-listener` container's logs
  - Verify that listener is running (`INFO:runner.messaging:Connected to message broker`)
  - Verify the worker is running (`celery@... ready`)
- Execute a task from Functionary's UI
  - Verify task logs show up in the `functionary-runner-listener` container
  - Verify task runs as expected


scott-taubman:
This looks good and works as expected in my tests.

One minor thing is that it feels like the top level listener.py and worker.py belong somewhere in the runner module hierarchy.  I don't feel too strongly about where, but two possible suggestions:

* Create a runner/process folder and put them in there
* Just put the two classes in the runner/__init__.py

I'm open to other ideas too.  They just feel slightly out of place sitting at the same level as the runner.py that spawns the two processes.

xeonneon:
- Moved the `Worker` and `Listener` classes to `runner/__init__.py`
- Removed `worker.py` and `listener.py`
- Rebased with main

I have noticed that occasionally, the worker and listener processes will start, but when inspecting the container's logs, it doesn't appear the processes are running. When this does happen, I don't see any messages being logged or any errors being raised. Restarting the container fixes the issue, but I'm not sure why the issue is inconsistent. I might need to add some more logging statements to the listener and worker to pinpoint what the issue could be before we merge this.

scott-taubman:
> I have noticed that occasionally, the worker and listener processes will start, but when inspecting the container's logs, it doesn't appear the processes are running. When this does happen, I don't see any messages being logged or any exceptions being raised. Restarting the container fixes the issue, but I'm not sure why the issue is inconsistent. I might need to add some more logging statements to the listener and worker to pinpoint what the issue could be before we merge this.

I ran into this as well.  I'm fairly convinced that this is actually a problem with the way the debugging is setup in the container entrypoint.  Using debugpy as the entrypoint process was really a holdover from an alternate path that was explored for the debugging workflow and was left in there just because it theoretically gave a path to easily attaching to the container process for debugging.  There are other ways to achieve that, the current entrypoint is kind of just confusing, and now it also appears to not be multiprocessing friendly, so I feel inclined to just simplify the entrypoint process.

I did also notice that debugging natively (not through the container) also doesn't work, seemingly because the implicit join() behavior that you're relying on doesn't play nice in the debugger context?  I'm not sure on that one but it seemed like the process was just exiting rather than waiting for the children.  So likely an explicit join is needed anyway.

Since outside of the debugging context this PR seems fine, I'm going to merge it and then separately address the debug workflow related issues, including updating the vscode tasks.

PR #105, opened 2022-10-17
#92 - Runner now uses filesystem broker instead of redis

### LINKED ISSUE #92: Remove redis dependency from the runner
We should explore using something like the filesystem transport since the runner's work queue is completely internal to itself and we really don't need an external broker for that.

Filesystem transport info: https://docs.celeryq.dev/projects/kombu/en/latest/reference/kombu.transport.filesystem.html
Simple example: https://www.distributedpython.com/2018/07/03/simple-celery-setup/

xeonneon:
Closes #92 

This PR removes the Redis dependency from the Runner by using the Filesystem broker for Celery.

## Introduced Changes
- Celery App for the Runner uses the Filesystem broker instead of the Redis broker

## Testing
- Start Functionary
- Inspect container logs for `functionary_runner`. In the config section of the Celery startup message, it should say Celery is using `filesystem://localhost//` for the `transport` variable
  - You should also see Celery repeatedly checking for new tasks every second
- Run a Functionary function to verify function gets executed as expected

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -3,6 +3,7 @@ version: "3.5"
 x-environment:
   &environment
   DEBUG: TRUE
+  BROKER_WORKDIR: /tmp
   DJANGO_SECRET_KEY: ${DJANGO_SECRET_KEY:-supersecret}
   DJANGO_SETTINGS_MODULE: functionary.settings.debug
   LOG_LEVEL: DEBUG
@@ -118,7 +119,6 @@ services:
       - /var/run/docker.sock:/var/run/docker.sock
     depends_on:
       - rabbitmq
-      - redis
   database:
     image: postgres:latest
     container_name: functionary-database
--- a/runner/runner/celery.py
+++ b/runner/runner/celery.py
@@ -11,8 +11,9 @@ from logging.config import dictConfig
 from celery import Celery
 from celery.signals import setup_logging
 
-REDIS_HOST = os.getenv("REDIS_HOST", "localhost")
-REDIS_PORT = os.getenv("REDIS_PORT", "6379")
+BROKER_WORKDIR = os.getenv("BROKER_WORKDIR", "/tmp")
+BROKER_WORKDIR_PATH = os.path.join(BROKER_WORKDIR, "broker")
+
 LOGGING = {
     "version": 1,
     "disable_existing_loggers": False,
@@ -38,10 +39,24 @@ LOGGING = {
     },
 }
 
+
 app = Celery(
     "runner",
-    broker=f"redis://{REDIS_HOST}:{REDIS_PORT}",
     include=["runner.handlers"],
+)
+
+app.conf.update(
+    {
+        "broker_url": "filesystem://",
+        "broker_transport_options": {
+            "data_folder_in": BROKER_WORKDIR_PATH,
+            "data_folder_out": BROKER_WORKDIR_PATH,
+        },
+        "result_persistent": False,
+        "task_serializer": "json",
+        "result_serializer": "json",
+        "accept_content": ["json"],
+    }
 )
 
 
--- a/runner/.vscode/.debug.env
+++ b/runner/.vscode/.debug.env
@@ -1,3 +1,4 @@
 LOG_LEVEL = "DEBUG"
 RABBITMQ_USER = "bugs"
 RABBITMQ_PASSWORD = "wascallywabbit"
+BROKER_WORKDIR = "/tmp"
--- a/runner/runner.py
+++ b/runner/runner.py
@@ -8,14 +8,25 @@ logging.basicConfig(level=LOG_LEVEL)
 logging.getLogger("pika").propagate = False
 
 
-def spawn_listener():
+def setup_broker_dir() -> None:
+    broker_workdir = os.getenv("BROKER_WORKDIR", "/tmp")
+    broker_workdir_path = os.path.join(broker_workdir, "broker")
+    try:
+        if not os.path.exists(broker_workdir_path):
+            os.makedirs(os.path.join(broker_workdir_path))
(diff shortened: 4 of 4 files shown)


docker/docker-compose.yml
```diff
@@ -181,3 +180,4 @@ volumes:
   redis-data:
   registry-data:
   postgresql-data:
+  celery-data:
```
scott-taubman, line 183: This isn't actually mounted to anything.  I assume it's supposed to be mounted up to `/var/lib/broker`.

scott-taubman (changes requested):

runner/runner/celery.py
```diff
@@ -38,12 +36,27 @@
...
     "runner",
-    broker=f"redis://{REDIS_HOST}:{REDIS_PORT}",
     include=["runner.handlers"],
 )
 
+app.conf.update(
+    {
+        "broker_url": "filesystem://",
+        "broker_transport_options": {
+            "data_folder_in": "/var/lib/broker/data",
```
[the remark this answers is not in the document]
xeonneon: Added the `BROKER_WORKDIR` env var that the Runner will use to create the `broker` directory. By default, it makes the directory in `/tmp`

I added a try/except to catch permission errors when creating the directory, let me know if you think this is unnecessary.

docker/dev_docker.Dockerfile
```diff
@@ -16,6 +16,8 @@ RUN apt-get update && \
     make gcc python3-dev libpq-dev \
     # packages for docker install
     ca-certificates curl gnupg lsb-release && \
+    mkdir -p /var/lib/broker/data && \
```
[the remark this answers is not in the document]
xeonneon: Resolved

docker/docker-compose.yml
xeonneon: Resolved. Remnant from when I made the builder use the filesystem broker

runner/runner/celery.py
```diff
@@ -38,12 +36,27 @@
...
+            "data_folder_out": "/var/lib/broker/data",
+            "data_folder_processed": "/var/lib/broker/processed",
```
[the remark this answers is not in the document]
xeonneon: If I understood correctly, I just removed the `data_folder_processed` option, and now the `data_folder_in/out` just points to the root of the `broker` directory.

runner/runner/celery.py
```diff
@@ -11,8 +11,17 @@
 from celery import Celery
 from celery.signals import setup_logging
 
-REDIS_HOST = os.getenv("REDIS_HOST", "localhost")
-REDIS_PORT = os.getenv("REDIS_PORT", "6379")
+broker_workdir = os.getenv("BROKER_WORKDIR", "/tmp")
+
+# Make subdirectory for filesystem broker
+try:
```
scott-taubman, line 17: Doing the directory creation here is problematic in that it makes the mere act of importing from this file something that has side effects.  This is almost always undesirable and frequently confusing.  So for instance, let's say I do the following:

```python
from runner.celery import app
```

Just by doing that, I've created a directory somewhere, assuming I have access to the `BROKER_WORKDIR`.  And if I don't have access, I get the PermissionError informing me that it failed to create the directory, at which point I'm like "geez I was just trying to import something".

There are various ways to refactor this that would avoid this problem.  An easy one might be to just have this portion of the code assume that the BROKER_WORKDIR is something that exists and is accessible.  Then have the main process code, the `runner.py` handle the check and possible creation that you have here.  Maybe just add a `setup_broker` method that gets called prior to calling `spawn_listener` and `spawn_worker`.  That's one option of many.  The point is just to avoid this import side-effect thing.

runner/runner/celery.py
```diff
@@ -11,8 +11,17 @@
...
+    if not os.path.exists(os.path.join(broker_workdir, "broker")):
+        os.makedirs(os.path.join(broker_workdir, "broker"))
+except PermissionError:
+    raise PermissionError(
+        f'You do not have permission to create {os.path.join(broker_workdir, "broker")}'
```
scott-taubman, line 22: A note for wherever you move this code to: it probably makes sense to set some other variable to the full path value and then just use that so that you're not calling `os.path.join` over and over again.  Something like:

```python
BROKER_WORKDIR = os.getenv("BROKER_WORKDIR", "/tmp")
BROKER_WORKDIR_DATA = os.path.join(BROKER_WORKDIR, "broker")

#  then later on
os.makdirs(BROKER_WORKDIR_DATA)  # etc, etc
```
It's less about the cost of doing the `os.path.join` and more about have fewer lines to update if we have to change the folder name for some reason.

Also worth pointing out is the all uppercase naming for module level constants.

scott-taubman (changes requested):

runner/runner/celery.py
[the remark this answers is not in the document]
scott-taubman: Yeah, what you did is fine.  I was originally suggesting that you still have a `/data` subdirectory so that you don't have to change the naming if we later decide that we need that `/processed` directory.  Basically, if the in and out folder is just `/tmp/broker` and we later want to add a processed directory, what is the path to that?  `/tmp/broker/data` makes the answer to that question obvious: `/tmp/broker/processed`.  I don't feel super strongly about it though and am fine with what you went with.

runner/runner/celery.py
xeonneon: Migrated the broker directory creation to `runner.py`.

I originally put the broker directory creation in the `celery.py` since I figured if you were ever going to import the app, you would want to ensure that the broker directory existed. 

I left the default parent directory of broker as `/tmp`, since it is r/w for everyone by default.

runner/runner/celery.py
xeonneon: Resolved

PR #108, opened 2022-10-17
#106: Remove milliseconds from datetime display

### LINKED ISSUE #106: Streamline date display in CLI results
Dates in the CLI results currently go well past seconds in granularity, which is overkill.  We should truncate it down to just seconds.

Current format: `2022-10-17T17:32:35.847934Z`
Desired format: `2022-10-17 17:32:35Z`

catattack05:
Closes #106 

Makes datetime objects present in cli tables come out formatted like '2022-10-13 19:36:18UTC'

### DIFF
--- a/cli/functionary/utils.py
+++ b/cli/functionary/utils.py
@@ -56,9 +56,7 @@ def _fix_datetime_display(value):
     Returns:
         value as a string representing datetime value without milliseconds
     """
-    value = datetime.datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%f%z").replace(
-        microsecond=0
-    )
+    value = datetime.datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%f%z")
     return value.strftime("%Y-%m-%d %H:%M:%S%Z")
 
 
--- a/cli/functionary/utils.py
+++ b/cli/functionary/utils.py
@@ -56,8 +56,8 @@ def _fix_datetime_display(value):
     Returns:
         value as a string representing datetime value without milliseconds
     """
-    value = datetime.datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%fZ").replace(
-        tzinfo=datetime.timezone.utc, microsecond=0
+    value = datetime.datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%f%z").replace(
+        microsecond=0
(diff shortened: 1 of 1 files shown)


cli/functionary/utils.py
```diff
@@ -45,6 +47,21 @@ def flatten(results, object_fields):
...
+def _fix_datetime_display(value):
+    """
+    Helper function for format_results to remove milliseconds from datetime
+
+    Args:
+        value: string representing datetime value
+    Returns:
+        value as a string representing datetime value without milliseconds
+    """
+    value = datetime.datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%fZ").replace(
```
scott-taubman, line 59: It's not obvious from the documentation, but I think using the %z format code will properly handle the `Z` designating UTC.  So this could just be:

```python
value = datetime.datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%f%z")
```

cli/functionary/utils.py
```diff
@@ -68,6 +85,8 @@ def format_results(results, title="", excluded_fields=[]):
                 continue
             if first_row:
                 table.add_column(key.capitalize())
+            if "_at" in key:
```
scott-taubman, line 88: Just to be extra safe, this should check that the key specifically **ends with** "_at".  Otherwise you run into something unexpected like:

```python
key = "favorite_attachment"
"_at" in key  # True
```

scott-taubman (changes requested):

cli/functionary/utils.py
```diff
@@ -45,6 +47,21 @@ def flatten(results, object_fields):
...
+def _fix_datetime_display(value):
+    """
+    Helper function for format_results to remove milliseconds from datetime
+
+    Args:
+        value: string representing datetime value
+    Returns:
+        value as a string representing datetime value without milliseconds
+    """
+    value = datetime.datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%f%z").replace(
```
scott-taubman, line 59: What's the reason replacing the microseconds?

catattack05: Was originally to remove the microseconds from the datetime object, but you're right that we don't need it now

scott-taubman (approved):

Issue #93, opened 2022-10-06
More useful output / return type handling


scott-taubman:
We want to be able to render output in a way that's most useful for the type of data that's being returned.  That is, json could be rendered nicely with interactive foldable stuffs.  Certain output might be best rendered as a table (json, csv, etc.).

We need to decide how to specify the return and output types in the package.yaml.  Then add that data to the appropriate places in either the model or the schema field of the model.  Then do something with that data in the ui.

PR #109, opened 2022-10-18
Add formatted output  - #93

kvig:
This adds CSV and JSON formatted output for Functionary results.

The `package.yaml` schema was updated to add the `return_type` and `output_format` parameters.
Supported values for `return_type` are:
- string
- json
- csv
Supported values for `output_format` are:
- json
- table
The task details page will do its best to coerce strings and json `result_type`'s to a JSON object.

In the case of an `output_format` set to `table`, csv and string `result_type`'s are assumed to be a string in csv format. These are parsed with the csv lib and the results set for rendering. For a json `result_type`, the output is assumed to be a JSON list or a JSON object with one key to a list of values, for example:
```json
[
  {
    "prop1":"value1",
    "prop2":"value2"
  },
  {
    "prop1":"value3",
    "prop2":"value4"
  }
]
```
or
```json
{
  "foos" : [
    {
      "prop1":"value1",
      "prop2":"value2"
    },
    {
      "prop1":"value3",
      "prop2":"value4"
    }
  ]
}
```

In all cases, if the conversion cannot be determined or there is an exception raised, the output will be the raw result and a message logged on the server.


scott-taubman (commented):
I made a few comments, but seeing the way this shaping out it makes me want to take a step back and rethink somewhat the target that we're after here.  More discussion to come.

functionary/ui/templates/tags/output_table.html
```diff
@@ -0,0 +1,18 @@
+{% if format == "table" %}
+    <table id="result_table" class="result_table">
```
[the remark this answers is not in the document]
kvig: Good call, completely forgot about Bulma

functionary/builder/utils.py
```diff
@@ -227,13 +238,19 @@ def _create_functions_from_definition(definitions, package: Package):
     db_functions = []
     for function_def in definitions:
         name = function_def.get("name")
+        ret_type = function_def.get("return_type")
+        if ret_type not in type_map:
+            raise ValueError(f"Incorrect return type for {name}")
```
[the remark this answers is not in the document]
kvig: Yes, that's definitely where it should be at

functionary/ui/templatetags/task_output.py
```diff
@@ -0,0 +1,134 @@
...
+                        r.append(row[key])
+                    res_data.append(r)
+
+                result = {"headers": headers, "data": res_data}
+            else:
+                raise ValueError("Unable to determine the structure of the JSON")
+        else:
+            raise ValueError("JSON data appears to be empty")
+    else:
+        raise ValueError(f"Unable to convert {result_type} to csv")
```
scott-taubman, line 79: Mostly just a nitpick for the sake of clarity, but both here and in the comment a few lines belong you refer to converting data to CSV.  It doesn't look like that's what you're doing necessarily.  You're organizing the data for purposes of displaying it as a table, right?

functionary/ui/templatetags/task_output.py
```diff
@@ -0,0 +1,134 @@
...
+                    e,
+                )
+
+        return f"<pre>{'<br/>'.join(the_task.result.splitlines())}</pre>"
+
+
+@register.tag("task_output")
+def do_task_output(parser, token):
+    """Convert the result of the task to the specified format."""
+    _, rest = token.contents.split(None, 1)
```
scott-taubman, line 133: I don't know what the right name for `rest`, but it's not obvious what data that refers to.  Is `tag_param` or something similar a better name?

functionary/ui/templates/tags/output_table.html
```diff
@@ -0,0 +1,20 @@
...
+                {% for row in result.data %}
+                    <tr>
+                        {% for col in row %}<td>{{ col }}</td>{% endfor %}
+                    </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+    </div>
+{% else %}
+    <pre>{{ result | linebreaksbr }}</pre>
```
scott-taubman, line 19: Seeing this in both templates strikes me as weird.  Should this just be its own `output_raw.html` template that gets rendered if the output_format is neither of json or html?

functionary/ui/templatetags/extras.py
```diff
@@ -10,7 +10,7 @@ def pretty_json(value):
     return json.dumps(value, indent=4)
 
 
-@register.filter
+@register.filter(is_safe=True)
 def input_type(value):
```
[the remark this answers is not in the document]
kvig: This is used for the parameters when the task form is displayed. Without this, the HTML inputs are essentially character fields allowing anything to be input vs numbers.

functionary/ui/templatetags/task_output.py
kvig: yes, that was a bad name

functionary/ui/templates/tags/output_table.html
kvig: That was an artifact from not having separate output templates. It's not accessible now so it has been removed.

scott-taubman (approved):

PR #116, opened 2022-10-21
110 task list performance

kvig:
These changes add `select_related()`s to the `get_queryset()`s used on the UI views to reduce excess database hits.

### DIFF
--- a/functionary/ui/views/functions.py
+++ b/functionary/ui/views/functions.py
@@ -30,10 +30,10 @@ class FunctionDetailView(PermissionedEnvironmentDetailView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        the_function = self.get_object()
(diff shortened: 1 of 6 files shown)


functionary/ui/views/functions.py
```diff
@@ -17,18 +17,23 @@
...
     model = Function
     environment_through_field = "package"
 
+    def get_queryset(self):
+        return super().get_queryset().select_related("package", "package__environment")
+
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        env = Environment.objects.get(id=self.request.session.get("environment_id"))
+        the_function = self.get_object()
```
scott-taubman, line 33: `function` is not a reserved word or builtin, so this can just be `function` rather than `the_function`.

scott-taubman (approved):

PR #119, opened 2022-10-25
Issue 115: Remove TeamId header from API

### LINKED ISSUE #115: Remove the X-Team-Id header from the API
Remove the X-Team-Id header from the API and all of the internal logic that tries to handle it.  This includes removing the concept of a "default" environment. 

NOTE: When a team is created, an environment should still get created for it.  That environment will just no longer be marked as the "default" environment.

catattack05:
Closes #115 

First draft PR to remove the X-Team-Id header from the API.

Changes I made:
-Remove X-Team-Id header from core/api/__init__
-Updated Environment model to no longer have a default Boolean field. 
-Updated Team serializer to no longer deal with default environments
-Update Team model's post creation hook to use Environment model's 'create_first_env' instead of 'create_default'. Basically just renamed 'create_default' and no longer set the environment made to be the default, although the created environment is still currently named default.
-Walk back some of the error handling I did a few tickets ago, so that we no longer have a specific custom error that deals with invalid teamid headers
-Removed mentions of X-Team-Id headers in comments
-Change test_task.py's Team Fixture to not try to search for default


functionary/core/models/team.py
```diff
@@ -24,4 +24,4 @@ def __str__(self):
     def post_create(self):
         """Post create hooks"""
         Environment = apps.get_model("core", "Environment")
-        Environment.create_default(self)
+        Environment.create_first_env(self)
```
scott-taubman, line 27: The original reason for the `create_default` method was that I didn't want the "Team" model to have to know the implementation details of declaring an Environment as the "default".  Not that those details were especially complex.

Given that there's nothing special about this initial environment that gets created now, I think it's okay to just create it here and delete get rid of the whole `create_default` / `create_first_env` method.  So this line could just become:

```suggestion
        Environment.objects.create(name="default", team=self)
```

functionary/core/models/environment.py
```diff
@@ -17,43 +17,36 @@ class Environment(models.Model):
...
-        default: when true, this environment will act as the default for a Team in
-                 cases where an environment is not specified.
     """
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     name = models.CharField(max_length=64)
     team = models.ForeignKey(
         to="Team", related_name="environments", on_delete=models.CASCADE, db_index=True
     )
-    default = models.BooleanField(default=False)
```
scott-taubman, line 29: Take a look at the file `core/fixtures/bootstrap.yaml`.  It has references to this field that will need to be deleted.  Once that's done, make sure you can successfully run the vscode task "functionary: apply migrations and load fixtures".

scott-taubman (changes requested):

scott-taubman (approved):

Issue #118, opened 2022-10-24
Backend support for task environment variables


kvig:
Create models to store team and environment level environment variables, with the ability to mark any given one as a "secret".

PR #120, opened 2022-10-27
Refs #118 - Add environment variables

kvig:
This change allows the admin interface to create environment variables for teams and environments. Using the `package.yaml`, the functions define the list of environment variables they need in order to run.

When getting the variables from the environment, any variables not found will be looked up in the associated team.

Example `package.yaml`:
```
version: 1.0
package:
  name: "test_package"
  summary: "This is for testing. Lets keep this going past 64 characters for this."
  language: "python"
  functions:
    - name: json_dict_output
      summary: Outputs a JSON object of a dict
      env_variables: [PATH, TEAM_ONLY_VAR]
      return_type: json
      output_format: table
      parameters:
        - name: unused
          description: Not used
          type: string
          default: "Hello Docker"
          required: false
```


functionary/core/models/environment_variable.py
```diff
@@ -0,0 +1,59 @@
...
+                name="environment_variable_only_one_team_or_environment",
+                check=(
+                    models.Q(team__isnull=True, environment__isnull=False)
+                    | models.Q(team__isnull=False, environment__isnull=True)
+                ),
+            ),
+        ]
+
+    def __str__(self):
+        return self.variable_name
```
scott-taubman, line 59: We have this problem elsewhere too, but I'll point out that there are likely to be variables across environments and teams that share a name and this current `__str__` makes all of them indistinguishable from each other.

scott-taubman (changes requested):

functionary/builder/utils.py
```diff
@@ -173,8 +173,11 @@ def build_package(build_id: UUID):
         build.save()
 
     logger.debug(f"Cleaning up remnants of build {build_id}")
-    docker_client.images.remove(image.id)
     shutil.rmtree(workdir)
+    # TODO this has been throwing can't remove image on my box,
+    # we can pass `force=True`, but is that what we want all to
+    # do all the time?
+    docker_client.images.remove(image.id)
```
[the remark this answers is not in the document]
kvig: I had been hitting this on basically every build working on this ticket, I'm not sure why it started acting up. Logging now enabled for it.

functionary/core/models/environment_variable.py
```diff
@@ -0,0 +1,59 @@
+""" Package model """
+import uuid
+
+from django.db import models
+
+
+class EnvironmentVariable(models.Model):
```
scott-taubman, line 7: Somewhat late thought here, but how would you feel about calling this just `Variable`?  And then on the function model and package.yaml the field would be `variables` rather than `env_variables`?  The reason for this suggestion is that even knowing exactly what this is, it still takes my mind a second each time to disassociate the "Environment" in "EnvironmentVariable" from a Functionary "Environment."  I feel like removing "Environment" from the model name here leaves less room for confusion.

scott-taubman: Also worth mentioning is that if we go that route, the `variable_name` field here probably just becomes `name`.

functionary/core/models/environment_variable.py
```diff
@@ -0,0 +1,59 @@
...
+        on_delete=models.CASCADE,
+        related_name="vars",
+        blank=True,
+        null=True,
+    )
+    team = models.ForeignKey(
+        "Team", on_delete=models.CASCADE, related_name="vars", blank=True, null=True
+    )
+
+    variable_name = models.CharField(max_length=64, blank=False)
```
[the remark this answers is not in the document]
kvig: Validator added

functionary/core/models/environment.py
```diff
@@ -29,3 +29,23 @@ class Meta:
 
     def __str__(self):
         return f"{self.team.name} - {self.name}"
+
+    def variables(self, variable_names=None):
```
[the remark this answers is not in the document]
kvig: Will add the docstring. 

This is used in the next step to get all environment variables visible from the environment. Optionally trimmed to just the values you pass in. Since the environment hides some from the Team, this seems like the best place to put the variable accessor.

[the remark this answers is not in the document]
scott-taubman: I think having a helper that gets you the full set of variables for the environment (its own, plus those inherited from the team) makes sense.  The part where you can pass in a list of variables to filter that down to feels a little out of scope for a model helper function to me.

And in any case, I think if you have this return a set answering that question can just be done via the set intersection.  For instance:

```python
In [14]: my_vars = {"two", "four"} # this is what you were proposing gets passed in as variable_names
In [15]: env_vars = {"one", "two", "three", "four"} # let's say this is what would have come back from calling variables()
In [16]: my_vars & env_vars
Out[16]: {'four', 'two'}
```

functionary/core/models/function.py
```diff
@@ -28,6 +28,7 @@ class Function(models.Model):
     display_name = models.CharField(max_length=64, null=True)
     summary = models.CharField(max_length=128, null=True)
     description = models.TextField(null=True)
+    env_variables = models.JSONField(default=list, blank=True, null=True)
```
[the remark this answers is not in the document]
kvig: fixed

functionary/core/models/environment_variable.py
kvig: renamed

functionary/core/models/function.py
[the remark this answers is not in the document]
kvig: validator added

functionary/core/models/environment_variable.py
kvig: added parent property as a shortcut to finding who owns the variable and use that for the name

functionary/core/models/environment.py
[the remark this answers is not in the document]
kvig: This function returns the variables, not just the names. You could do a comprehension when you get them all back but it seems silly to do that when you can just pass in the ones you want. User has similar methods for permissions.

functionary/core/models/variable.py
```diff
@@ -0,0 +1,69 @@
...
+class Variable(models.Model):
+    """An Environment Variable is set in the system and set in
+    the runtime environment for tasks as required by the function
+    in the package.yaml.
+
+    Attributes:
+        id: unique identifier (UUID)
+        environment: the environment that this variable belongs to
+        team: the team that this variable belongs to
+        variable_name: name of the environment variable to set
```
scott-taubman, line 22: variable_name --> name

functionary/core/models/function.py
```diff
@@ -28,6 +38,9 @@ class Function(models.Model):
     display_name = models.CharField(max_length=64, null=True)
     summary = models.CharField(max_length=128, null=True)
     description = models.TextField(null=True)
+    env_variables = models.JSONField(
```
scott-taubman, line 41: I vote to just call this `variables` here and in the package definition yaml too.

functionary/core/models/variable.py
kvig: fixed

scott-taubman (approved):

PR #122, opened 2022-10-27
#93 - Further refine results output formatting

### LINKED ISSUE #93: More useful output / return type handling

scott-taubman:
Closes #93 

This PR builds on the previous work for displaying nicely formatted JSON results and table output where available.  It has two main goals:

* To have the /tasks/{id}/result/ API show the results as the correct type, rather than always being stringified JSON.  Objects now return as objects, numbers as numbers, etc.
* To no longer require explicitly declaration of the output format up front on the function definition.  It was previously the case that when viewing the results you would be shown only a the raw JSON or a table, based on what output format had been declared.  Now you can toggle between both formats when appropriate like this:

### JSON Raw Output
image

### JSON Table Output
image

## Other Notes

* I removed the `output_format` field from the Function and Task model.  It turned out we can do everything we want to without it.
* I also added a permissions check to the unicorn view, as it was possible for data to be accessed through the unicorn endpoint by any logged in user before, regardless of their actual permissions.

## Testing Instructions

* Both the python and javascript templates had to be updated to support these changes.  Be sure to run the `./build.sh -p` from the templates folder to get the new versions of the templates in your registry.
* I added a simple "demo" example package that has useful functions for testing the UI rendering of JSON and text (including CSV formatted text).  Publish that package to an environment (**after publishing the new templates**) to ease your testing.
* Task the output_json and output_text functions with data that can and cannot be rendered as a table. The descriptions of each function explain what should generate table output.
* Retrieve the task results via the API as well and verify that the date type of the `result` property is what you expect based on the return type of the function.


kvig (commented):
All the changes look good. My only complaint, since you're refining the output, is that the delay function "has an output' but it's empty and it looks like the output is missing. Can you put a message when the output is blank?

scott-taubman:
> All the changes look good. My only complaint, since you're refining the output, is that the delay function "has an output' but it's empty and it looks like the output is missing. Can you put a message when the output is blank?

Added a message indicating that there is no result for a task.  I had to update the package templates again, so if you go to test it, make sure you rebuild the templates and then republish the package.

kvig (approved):
Looks good

PR #123, opened 2022-10-31
Issue 113: Adding BuildLog

### LINKED ISSUE #113: Record package build errors
During the package build process, we set the build status as "ERROR" when a build fails, but we do not actually record the error information anywhere.  Even in the case of a successful build it could be useful to be able to see the log output.

To capture the build output and error info we'll want to do the following:
- Create a `BuildLog` model (similar to the `TaskLog` model, but with a foreign key to a build rather than a task).
- In `functionary.builder.utils.build_package`, create the BuildLog entry in the appropriate places. Some things to consider:
  - The "places" in question are the two spots where `build.status` is set to `COMPLETE` or `ERROR` and then saved.
  - The build output should already be getting captured [here](https://github.com/beer-garden/functionary/blob/main/functionary/builder/utils.py#L154), we just need to persist that info in the `BuildLog` instance for the build.  NOTE: The `build_log` will come back as an iterator, not a blob of text.  Just something to be aware of.
  - The call to push the built image up to the registry (seen [here](https://github.com/beer-garden/functionary/blob/main/functionary/builder/utils.py#L160)) also returns a generator for the log.  We should capture that too and record it as part of the `BuildLog`.  So a single `BuildLog.log` will contain the output from both the build and the push.
  - In the event of an exception being raised (i.e. the build failed), treat the exception message (probably not the whole stack trace) as the log contents.
  - While we're in that specific exception handling block, let's update the `except Exception` to handle specific exceptions instead.  See the [official documentation](https://docker-py.readthedocs.io/en/stable/images.html) for what exceptions might get raised.

catattack05:
Closes #113 

Posting a PR so people can see where it's at. There should be a BuildLog created for each build you create visible in the Django admin page (with). Some things I would like feedback on:
-Right now I'm pretty much just grabbing all the lines spat out on push/build and sticking them into the log as is. Wasn't sure if we wanted them to look slightly nicer than that.
-The push results can either come back as a string or a generator, but I haven't actually seen them come back as a generator yet, so I did make a guesstimate about what that would look like.
-There are two relevant errors here (Docker's BuildError and APIError). I've got them in separate except blocks right now with an if block after that to mark build as error, save the build and log objects then return. Wasn't sure if there was a nicer way to format it and repeat minimal code. BuildError actually creates its own log that I'd like to attach to the model log, whereas APIError has response and explanation that I haven't attached yet. Those both seemed to be like things that might be useful, but was a little unsure (in my initial round of doing things I haven't seen these errors trigger yet)


functionary/builder/utils.py
```diff
@@ -105,6 +107,40 @@ def initiate_build(
...
+
+    Returns:
+        string with both build and push information in more readable format
+    """
+    build_str = ""
+    for line in build_results:
+        line_str = ""
+        for value in line.values():
+            line_str = line_str + str(value) + " "
+        build_str = build_str + line_str + "\n"
```
scott-taubman, line 127: It doesn't look like you need the newline here.  With it, you end up with a bunch of empty lines likes so:

```
Step 2/7 : ARG install_dir=/usr/src/app                                                                                                                                                                              
                                                                                                                                                                                                                     
                                                                                                                                                                                                                     
 ---> Using cache                                                                                                                                                                                                    
                                                                                                                                                                                                                     
 ---> a2b2693e4d1d                                                                                                                                                                                                   
                                                                                                          
Step 3/7 : COPY . $install_dir/
```

functionary/builder/utils.py
```diff
@@ -151,26 +187,32 @@ def build_package(build_id: UUID):
...
+            for func in db_functions:
+                func.save()
+            package.image_name = image_name
+            package.save()
+
+            build_log_message = _format_log_string(build_result, push_result)
+
+        log = BuildLog.objects.create(build=build, log=build_log_message)
         build.save()
+        log.save()
```
scott-taubman, line 215: You don't need this save, since the `BuildLog.objects.create` call does the save for you.  And since there's no use of `log` besides this, you can drop that assignment as well.

functionary/builder/utils.py
```diff
@@ -151,26 +187,32 @@ def build_package(build_id: UUID):
...
-        package.image_name = image_name
-        package.save()
-        build.status = Build.COMPLETE
+        if build.status == Build.COMPLETE:
+            for func in db_functions:
+                func.save()
+            package.image_name = image_name
+            package.save()
+
+            build_log_message = _format_log_string(build_result, push_result)
```
scott-taubman, line 211: `_format_log_string` is currently only called in the successful build path, meaning the logs are not properly formatted when there is an error.

functionary/builder/utils.py
```diff
@@ -105,6 +107,40 @@ def initiate_build(
...
+        push_results: generator object returned from docker push
+
+    Returns:
+        string with both build and push information in more readable format
+    """
+    build_str = ""
+    for line in build_results:
+        line_str = ""
+        for value in line.values():
+            line_str = line_str + str(value) + " "
```
scott-taubman, line 126: It looks like your intention is to have a space before each line, so just a heads up that as is you don't end up with a space before the first line.  You can fix that by injecting the space in the next line, similar to what you do with the `push_str`:

```python
line_str = line_str + str(value)
build_str = build_str + " " + line_str
```

scott-taubman (changes requested):

scott-taubman (approved):

PR #125, opened 2022-11-02
Refs 121 - Pass environment variables to runners

kvig:
The environment variables required by the functions are passed into the celery task and then added to the docker run command with the `environment=` parameter. A rudimentary filter on the output is run for any variable with protect enabled.


scott-taubman:
Other comments not related to the code:

* Some of the functionality seems like it could be unit testable.  For example:
  * the right environment variables get included in the published tasking message (only those listed on the function, the right variables based on team / environment variable inheritance, etc).
  * verifying that the task log output gets properly masked based on the protected variables.
* We should update the `cli/functionary/templates/package.yaml` with an example and description of the `variables` parameter.
* Add a function to the `demo` package that can be used to test / demonstrate the variable functionality.  Here's the function I wrote to test the PR, which you could probably just copy and past into the demo package:
  ```python
  # NOTE: dummy is just there because having no parameters causes issues now. 
  #       I made a separate note to fix that bug
  def variables(dummy):
      """Print environment variables that got passed through"""
      vars = []
      for name, value in os.environ.items():
          vars.append(f"{name}: {value}")

      output = "\n".join(vars)

      print(output)
      return output
  ```

runner/runner/handlers.py
```diff
@@ -35,22 +37,30 @@ def run_task(_=None, task=None):
...
+    try:
+        container = docker_client.containers.run(
+            package,
+            auto_remove=False,
+            detach=True,
+            command=run_command,
+            environment=variables,
+        )
+    except DockerException:
+        return (1, "Unable to execute function", "")
```
kvig, line 63: Had to add this because I keep making a `PATH` variable and that blows away the containers `PATH`. This results in the container failing to start and you get 0 feedback anywhere. It's not the best for now, but it at least doesn't leave the task in an incorrect state.

scott-taubman: This seems fine.  I would include the exception message in your output though since I'm guessing the additional context would be helpful:

```python
except DockerException as exc:
    return (1, f"Unable to execute function. Encountered error: {exc}", "null")
```

Also the "null" instead of empty string because that's what would normally come back if nothing is returned.

PR #130, opened 2022-11-07
Issue 124: Add build viewing to UI

### LINKED ISSUE #124: Display Package Builds in the UI
The build status is currently only available via the CLI, with the build logs not being available anywhere.  We want to add a page in the UI to make the builds viewable.

* Add a "Builds" menu item below "Functions"
* The Builds page shows a simple table listing the builds and their status.
* Clicking an entry on the table brings you to the individual Build details, where you can see additional details that were not in the list view, such as the log output.

catattack05:
Closes #124 

Added two new templates: functionary/ui/templates/builder/build_list.html and functionary/ui/templates/builder/build_detail.html. Two new corresponding views to go with them in functionary/ui/views/builds.py. Updated functionary/ui/urls.py with routes to the new views.

The one minor thing I saw to add was making the newline characters in the log print out. The editor didn't seem to like it style wise if I did it using inline styles on the html page, but the css file involved here looks like a ton of preset stuff.

### DIFF
--- a/functionary/ui/templates/builder/build_detail.html
+++ b/functionary/ui/templates/builder/build_detail.html
@@ -7,7 +7,7 @@
                     <a href="{% url 'ui:build-list' %}">Build List</a>
                 </li>
                 <li class="is-active">
-                    <a href="#">{{ object.id }}</a>
+                    <a href="#">{{ object.package }}</a>
                 </li>
             </ul>
         </nav>
@@ -16,7 +16,7 @@
         <div class="block">
             <h1 class="title is-1">
                 <span class="icon"><i class="fa fa-wrench"></i></span>&nbsp;&nbsp;
-                {{ object.id }}
+                {{ object.package }}
             </h1>
         </div>
         <div class="column has-addons">
@@ -44,22 +44,16 @@
                     <span class="ml-4">{{ object.status }}</span>
                 </div>
             {% endif %}
-            {% if object.environment %}
+            {% if object.id %}
                 <div class="field">
-                    <label class="label" for="desc">Environment:</label>
-                    <span class="ml-4">{{ object.environment }}</span>
-                </div>
-            {% endif %}
-            {% if object.package %}
-                <div class="field">
-                    <label class="label" for="desc">Package:</label>
-                    <span class="ml-4">{{ object.package }}</span>
+                    <label class="label" for="desc">Build Id:</label>
+                    <span class="ml-4">{{ object.id }}</span>
                 </div>
             {% endif %}
             {% if build_log.log %}
                 <div class="field">
                     <label class="label" for="desc">Build Log:</label>
-                    <span class="ml-4">{{ build_log.log }}</span>
+                    <pre class="ml-4">{{ build_log.log }}</pre>
                 </div>
             {% endif %}
         </div>
--- a/functionary/ui/templates/builder/build_detail.html
+++ b/functionary/ui/templates/builder/build_detail.html
@@ -1,0 +1,67 @@
+{% extends "base.html" %}
+{% block content %}
+    <div class="column is-full">
+        <nav class="breadcrumb has-arrow-separator" aria-label="breadcrumbs">
+            <ul>
+                <li>
+                    <a href="{% url 'ui:build-list' %}">Build List</a>
+                </li>
+                <li class="is-active">
+                    <a href="#">{{ object.id }}</a>
+                </li>
+            </ul>
+        </nav>
+    </div>
+    <div class="column">
+        <div class="block">
+            <h1 class="title is-1">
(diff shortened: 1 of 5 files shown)


catattack05:
As another random note, I just picked a wrench icon to represent the build for now. That can be changed pretty easily, was just trying to look for something vaguely 'building' related

functionary/ui/views/builds.py
```diff
@@ -0,0 +1,20 @@
...
+    model = Build
+    order_by_fields = ["id"]
+
+
+class BuildDetailView(PermissionedEnvironmentDetailView):
+    model = Build
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["build_log"] = BuildLog.objects.filter(build=context["object"])[0]
```
scott-taubman, line 19: You can get the BuildLog through the Build object:

```python
context["build_log"] = context["object"].buildlog
```

Also, regardless of which way you do it, the build log will not exist prior to the build completing, so you'll have to handle that here.  One option is to catch ObjectDoesNotExist (from django.core.exceptions):

```python
try:
    context["build_log"] = context["object"].buildlog
except ObjectDoesNotExist:
    context["build_log"] = None

File: functionary/ui/templates/builder/build_list.html [hunk 2]
```diff
@@ -0,0 +1,25 @@
...
+                    <div class="card-content">
+                        <div class="media">
+                            <div class="media-left">
+                                <span class="icon is-large">
+                                    <i class="fa fa-2x fa-wrench"></i>
+                                </span>
+                            </div>
+                            <div class="media-content">
+                                <a href="{% url 'ui:build-detail' build.id %}">
+                                    <p class="title is-4">{{ build.id }}</p>
```
scott-taubman, line 16: In places where you're currently displaying the `build.id`, I think it's probably more useful to show the `build.package.name`.  That's the thing that is going to be recognizable to people.  I think that's here and on the build detail page as well.

Also, I think there's other information we could display in the list view that would be useful:

* created_at
* creator
* status

Take a look at how the task list view is formatted.  Something similar would probably work here.

functionary/ui/templates/builder/build_detail.html
```diff
@@ -0,0 +1,67 @@
...
+            {% if object.environment %}
+                <div class="field">
+                    <label class="label" for="desc">Environment:</label>
+                    <span class="ml-4">{{ object.environment }}</span>
+                </div>
+            {% endif %}
+            {% if object.package %}
+                <div class="field">
+                    <label class="label" for="desc">Package:</label>
+                    <span class="ml-4">{{ object.package }}</span>
```
scott-taubman, line 56: This goes with my other comment of showing the package name instead of the build id in most places...This is the spot where I'd include the build id, since the main heading will have the package name already.

functionary/ui/templates/builder/build_detail.html
```diff
@@ -0,0 +1,67 @@
...
+            {% endif %}
+            {% if object.status %}
+                <div class="field">
+                    <label class="label" for="desc">Status:</label>
+                    <span class="ml-4">{{ object.status }}</span>
+                </div>
+            {% endif %}
+            {% if object.environment %}
+                <div class="field">
+                    <label class="label" for="desc">Environment:</label>
```
scott-taubman, line 49: We don't need to display the environment here, since you are only ever looking at the builds for a single environment.  Basically you select an environment up front and then all data you are looking at is for that environment, so displaying the environment again anywhere ends up being redundant.  This redundancy shows up in other places currently and needs to be cleaned up, but we may as well clean it up here now.

functionary/ui/templates/builder/build_detail.html
```diff
@@ -0,0 +1,67 @@
...
+            {% if object.package %}
+                <div class="field">
+                    <label class="label" for="desc">Package:</label>
+                    <span class="ml-4">{{ object.package }}</span>
+                </div>
+            {% endif %}
+            {% if build_log.log %}
+                <div class="field">
+                    <label class="label" for="desc">Build Log:</label>
+                    <span class="ml-4">{{ build_log.log }}</span>
```
scott-taubman, line 62: If you make this a `pre` instead of a `span` it will respect the newlines and the log is much more legible.

functionary/ui/views/builds.py
```diff
@@ -0,0 +1,20 @@
...
+
+from .view_base import (
+    PermissionedEnvironmentDetailView,
+    PermissionedEnvironmentListView,
+)
+
+
+class BuildListView(PermissionedEnvironmentListView):
+    model = Build
+    order_by_fields = ["id"]
```
scott-taubman, line 11: Because we're using uuids, sorting by id ends up not being especially meaningful.  I think we want this to instead by:

```python
order_by_fields = ["-created_at"]
```

scott-taubman (approved):

PR #132, opened 2022-11-15
#128 - Fix issue tasking functions with no parameters

### LINKED ISSUE #128: Functions without parameters cannot be executed
It is not currently possible to execute functions with no parameters through the UI.

scott-taubman:
Closes #128 

This PR fixes an issue that prevented tasking a function that takes no parameters via the UI.

## Test Instructions
* Publish the updated demo package
* Task the "variables" function, which now takes now parameters (rather than a dummy parameter that existed as a workaround)

### DIFF
--- a/examples/demo/functions.py
+++ b/examples/demo/functions.py
@@ -18,7 +18,7 @@ def long_running_process(duration: int = 60):
     time.sleep(duration)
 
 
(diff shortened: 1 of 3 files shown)


kvig (approved):
Works for me.

PR #133, opened 2022-11-17
Add variables to the UI

kvig:
This PR adds the ability to show the missing variables that functions require when attempting to execute a function.

It also allows a user with the proper permissions to add/update/delete the variables. The Environment and Team detail pages have the variables listed on them where a user can maintain the Variables for the system.


functionary/ui/templates/core/team_detail.html
```diff
@@ -1,42 +1,53 @@
...
+                    {% for u in users %}
+                        <li class="block ml-4">
+                            <p>
+                                {{ u.first_name }} {{ u.last_name }} <span class="has-text-grey-light">({{ u.username }})</span>
+                            </p>
+                        </li>
+                    {% endfor %}
+                </ul>
+            </div>
+            {% include 'partials/variables.html' with parent_object=team parent_id=team.id %}
```
kvig, line 50: This was the only line that was actually changed in the file

functionary/ui/templates/core/package_list.html
```diff
@@ -21,7 +21,11 @@
                                         {% endif %}
                                     </p>
                                 </a>
-                                {% if package.summary %}<p class="subtitle is-6">{{ package.summary }}</p>{% endif %}
+                                {% if package.summary %}
```
kvig, line 24: Default to showing the summary if available, otherwise try show the description.

functionary/ui/templates/unicorn/task_detail.html
```diff
@@ -1,10 +1,12 @@
 <div class="column"
-     {% if should_refresh %} unicorn:model="task" unicorn:poll-5000="refresh_task"{% endif %}>
+     {% if should_refresh %} unicorn:model="task" unicorn:poll-5000="refresh_task"{% endif %}
```
kvig, line 2: This code is used to target partial updates to the UI instead of rendering the whole thing

functionary/ui/templates/forms/variable_form.html
```diff
@@ -0,0 +1,62 @@
+{% load widget_tweaks %}
+{# djlint: off #}
+{% if add %}
+    <div class="column is-three-quarters add_variable_form" hx-target="this" hx-swap="outerHTML">
+{% else %}
+    <td colspan="3" class="add_variable_form" hx-target="this" hx-swap="outerHTML">
+{% endif %}
+        <form method="post">
+            {% csrf_token %}
+            <input type="hidden" name="parent_id" value="{{ parent_id }}"/>
```
scott-taubman, line 10: Where does this hidden field stuff get used?  It looks like the parent_id is actually read from the kwargs and not the form (which feels like the correct behavior), and then the looping over `form.hidden_fields` seems like it would always be looping over an empty list.

functionary/ui/templates/core/function_detail.html
```diff
@@ -69,6 +57,34 @@ <h1 class="title is-1">
...
+                    method="post"
+                    action="{% url 'ui:function-execute' %}">
+                {% csrf_token %}
+                <input type="hidden" name="function_id" value="{{ function.id }}"/>
+                <div class="column has-addons">
+                    {{ form.non_field_errors }}
+                    {{ form }}
+                    {% if missing_variables %}
+                        <div class="field ml-4 missing-variables has-background-danger-light has-text-danger">
+                            <span>Tasking this function may not work as expected, the following variables are missing:</span>
```
scott-taubman, line 72: Super nitpick: I think the comma needs to be a semicolon for this to read correctly.

functionary/ui/forms/variables.py
```diff
@@ -0,0 +1,56 @@
...
+                    name=var_name, team_id=self.parent_id
+                ).exists()
+            ):
+                raise ValidationError(
+                    f"A Variable named {var_name} already exists for {self.parent_id}"
+                )
+
+        env = cleaned_data.get("environment", None)
+        team = cleaned_data.get("team", None)
+        if env and team:
```
scott-taubman, line 51: This mutual exclusivity check is also handled by the model and having both set appears to me to be a condition that is impossible for a user to actually induce, since only a single id (the "parent_id") is provided.  It seems safe to me to drop this whole check.

functionary/ui/forms/variables.py
```diff
@@ -0,0 +1,56 @@
...
+
+            if not self.instance and (
+                Variable.objects.filter(
+                    name=var_name, environment_id=self.parent_id
+                ).exists()
+                or Variable.objects.filter(
+                    name=var_name, team_id=self.parent_id
+                ).exists()
+            ):
+                raise ValidationError(
```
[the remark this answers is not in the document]
kvig: This was a holdover from a previous implementation, removed.

functionary/ui/forms/variables.py
kvig: I think I induced this by hacking part of the request process. Removed the whole clean function.

functionary/ui/templates/core/function_detail.html
kvig: Done

functionary/ui/templates/forms/variable_form.html
kvig: That appears to be correct. Removed.

functionary/ui/views/variables.py
```diff
@@ -0,0 +1,152 @@
...
+
+    if request.user.has_perm(Permission.VARIABLE_DELETE, variable):
+        variable.delete()
+        return HttpResponse("")
+
+    return HttpResponseForbidden(f"Permission denied deleting {variable.name}")
+
+
+class VariableView(LoginRequiredMixin, UserPassesTestMixin, View):
+    form_class = VariableForm
```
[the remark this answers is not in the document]
kvig: That would explain some things :D 

Removed

PR #135, opened 2022-11-22
#134 - Workflow Backend

### LINKED ISSUE #134: Task sets (run tasks in series, with output of one acting as input for the next)
We want to support sequential series of tasking, where the output of one task could feed the input of a subsequent task.  This issue is for the backend portion of that and should include:

* Models for defining the Task Sets
* Models for tracking the execution of a given Task Set
* Updates to the task results processing to execute the next task in the sequence

scott-taubman:
Closes #134 

This PR implements the backend portion of "Workflows" (previously referred to as Task Sets).  A Workflow allows a series of functions to be executed sequentially, with the input of Tasks optionally being derived from either a shared set of parameters for the Workflow, or the results of Tasks from earlier in the Workflow.

The implementation of this is almost entirely in 4 new models that are introduced.  There is a tiny bit of code added to the `record_tasking_results` function that simply checks if a task is part of a workflow and fires off the next step as appropriate.

The 4 new models are:

* `Workflow` - The top level record that defines things like name, description, creator, and parameters.
* `WorkflowStep` - An individual step in a workflow.  When a workflow is run, each WorkflowStep will spawn a Task.
* `WorkflowRun` - An actual execution of a Workflow
* `WorkflowRunStep` - An execution of a Step within the WorkflowRun. Used for relationship tracking so that we can get from a Task all the way back to the Workflow it derived from.

Things not yet implemented:
* User-facing creation / management (UI and/or API)
* WorkflowRun.parameter validation against Workflow.schema. The relationship is similar to Function.schema --> Task.parmeters.
* Helpers for adding steps to a workflow (to handle updating the "next" pointers appropriately).
* As of now there are no unit tests.  There are some things that I want to build on (UI for instance) and get confidence that this design is going to hold up before I spend time writing the tests.

## Test Instructions

There is no UI or API currently, so this has to be tested in a django shell.

* Run the migrations and load fixtures
* A decent usage example looks like this:

```python
from core.models import *

admin = User.objects.get(username="admin")
ff_dev = Environment.objects.get(name="dev")
add_function = Function.objects.get(package__environment=ff_dev, name="add")
lrp_function = Function.objects.get(package__environment=ff_dev, name="long_running_process")


# A workflow that establishes some parameters
workflow = Workflow.objects.create(
    environment=ff_dev, name="DemoWorkflow", description="Show off workflows", creator=admin, schema={}
)

# Step that uses the Workflow's parameters
ws1 = WorkflowStep.objects.create(
    workflow=workflow,
    function=add_function,
    name="add_function_1",
    parameter_template='{"a": 1, "b": {{parameters.start_at}}}',
)

# Step that uses the result from previous step
ws2 = WorkflowStep.objects.create(
    workflow=workflow,
    function=add_function,
    name="add_function_2",
    parameter_template='{"a": 1, "b": {{add_function_1.result}}}',
)

ws1.next = ws2
ws1.save()

# Step that uses results from both previous steps
ws3 = WorkflowStep.objects.create(
    workflow=workflow,
    function=add_function,
    name="add_function_3",
    parameter_template='{"a": {{add_function_1.result}}, "b": {{add_function_2.result}}}',
)

ws2.next = ws3
ws2.save()

# Step that uses the results of the last add to do something that's not an add
ws4 = WorkflowStep.objects.create(
    workflow=workflow,
    function=lrp_function,
    name="lrp_function",
    parameter_template='{"duration": {{add_function_3.result}}}',
)

ws3.next = ws4
ws3.save()

run = WorkflowRun.objects.create(
    workflow=workflow,
    environment=workflow.environment,
    creator=admin,
    parameters={"start_at": 10}
)

# This will fire off the first task. Subsequent tasks should fire off as previous task results get processed.
run.execute()
```

You should see tasking get generated as the various steps resolve, with the parameters getting set appropriately.  Feel free to play around with building other types of workflows.


functionary/core/models/workflow_step.py
```diff
@@ -0,0 +1,95 @@
...
+from django.template import Context, Template
+
+from core.models import Task, Workflow, WorkflowRunStep
+
+if TYPE_CHECKING:
+    from core.models import WorkflowRun
+
+
+VALID_STEP_NAME = RegexValidator(
+    regex="^[a-zA-Z0-9_]+$",
```
xeonneon, line 16: I think this could just be simplified to `^\w+$`

functionary/core/models/workflow_run.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def error(self) -> None:
+        """Set the WorkflowRun status to ERROR"""
+        self._update_status(Task.ERROR)
+
+    def in_progress(self) -> None:
+        """Set the WorkflowRun status to IN_PROGRESS"""
+        self._update_status(Task.IN_PROGRESS)
+
+    def execute_next_step(self) -> Optional[Task]:
```
kvig, line 118: Should we prevent this from running if the current step hasn't finished yet? I tried running this by hand and my setup was busted so the first Task errored out. When I tried to run this on the Django shell, the json.loads in `WorkflowStep::_get_parameters()` failed because it was referencing the previous steps output.

functionary/core/models/workflow.py
```diff
@@ -0,0 +1,46 @@
...
+        description: details about the Workflow
+        schema: describes the parameters that runs of this Workflow may provide
+        creator: the user that initiated the task
+        created_at: task creation timestamp
+        updated_at: task updated timestamp
+    """
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    environment = models.ForeignKey(to="Environment", on_delete=models.CASCADE)
+    name = models.CharField(max_length=64)
```
kvig, line 24: Does this need a validator like the WorkflowStep's name?

functionary/core/models/workflow_step.py
kvig: Agreed, this wasn't used in the Variables since we wanted those to be all uppercase.

kvig (changes requested):
Good start, just a couple things to fix.

xeonneon:
Is there a reason behind using a sequence number to identify what `WorkflowStep` comes next, versus just a nullable Foreign Key reference to the next `WorkflowStep`, kinda like a linked list? I figured this would simplify the logic in `workflow_run.py:execute_next_step()`, since you would just execute the next `WorkflowStep` if it's not null

functionary/core/utils/tasking.py
```diff
@@ -132,3 +136,14 @@ def _update_task_status(task: Task, status: int) -> None:
 
     if task.scheduled_task is not None and status == "ERROR":
         task.scheduled_task.error()
+
+
+def _handle_workflow_run(workflow_run: WorkflowRun, task: Task) -> None:
+    """Start the next task for a WorkflowRun or update its status as appropriate"""
+    if task.status == Task.COMPLETE:
+        next_task = workflow_run.execute_next_step()
```
xeonneon, line 144: This function could be simplified to just being a `match` case statement, which may improve readability in the future if we add more task statuses.  

The `next_task` conditional could use the walrus operator so you don't need the separate conditional block if `next_task` is `None`

scott-taubman:
> Is there a reason behind using a sequence number to identify what `WorkflowStep` comes next, versus just a nullable Foreign Key reference to the next `WorkflowStep`, kinda like a linked list? I figured this would simplify the logic in `workflow_run.py:execute_next_step()`, since you would just execute the next `WorkflowStep` if it's not null

Yeah, I had started to mull this idea over too and it probably is the more straightforward approach in terms of figuring out what the next step is.  

I think one of my concerns is that it becomes kind of ugly to order the steps for purposes of visualizing the workflow.  Having a sequence number makes that part of it trivial because you just order the steps by the sequence number.  And the logic for finding the next step is also trivial because it's just whatever the lowest remaining sequence number is.  The complication comes with when you want to reorder the steps or inject new ones.  The sequencing could potentially be messy there and I was already a little concerned with what that was going to look like.

I'll switch to the linked-list style and see how that feels.  The ordering bit will be a little messier, but shouldn't really be that bad.

functionary/core/models/workflow_step.py
```diff
@@ -0,0 +1,95 @@
...
+
+        resolved_parameters = (
+            Template(self.parameter_template or "{}")
+            .render(context)
+            .replace("&quot;", '"')
+        )
+
+        return json.loads(resolved_parameters)
+
+    def execute(self, workflow_run: "WorkflowRun") -> Task:
```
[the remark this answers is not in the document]
scott-taubman: I think it's the type hint you are asking about and yes, it's correct.  I'm guessing you confused it for a default because of the quotes, which are necessary because `WorkflowRun` is only imported when type checking.

functionary/core/models/workflow_step.py
```diff
@@ -0,0 +1,95 @@
...
+        """Executes a Task base on this element's function and parameters
```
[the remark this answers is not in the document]
scott-taubman: Yes on both counts.  A previous version of this had some pretty terrible naming, in which "steps" were referred to as "elements."  I think I've updated all of those references now.

functionary/core/models/workflow_run.py
```diff
@@ -0,0 +1,145 @@
...
+    def get_context(self) -> Context:
+        """Generates a context for resolving tasking parameters.
+
+        Returns:
+            A Context containing data from all WorkflowRunSteps that have
+            occurred for this WorkflowRun
+        """
+
+        # Malleable object class for storing context properties
+        class Object:
```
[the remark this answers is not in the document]
scott-taubman: I'm not sure what I originally thought the templating engine wasn't going to like and lead to me creating this wrapper object, but it's not necessary.  I switched this to over to just being a dict.

functionary/core/models/workflow_run.py
```diff
@@ -0,0 +1,145 @@
...
+        return Context(context)
+
+    def _update_status(self, status: str) -> None:
+        if self.status != status:
+            self.status = status
+            self.save()
+
+    def complete(self) -> None:
+        """Set the WorkflowRun status to COMPLETE"""
+        self._update_status(Task.COMPLETE)
```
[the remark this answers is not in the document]
scott-taubman: I contemplated this as well.  I'm not really sure what another logical place is and the workflow statuses are somewhat purposefully just using the same statuses as individual tasks, so I'm kind of inclined to leave them on the Task model, at least for now.

functionary/core/models/workflow_run.py
```diff
@@ -0,0 +1,145 @@
...
+        steps = self.steps.all()
+
+        if not steps.exists():
+            max_sequence = 0
+        else:
+            max_sequence = steps.aggregate(models.Max("workflow_step__sequence"))[
+                "workflow_step__sequence__max"
+            ]
+
+        # This intentionally avoids specifying what the exact sequence number is.
```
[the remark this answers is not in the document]
scott-taubman: I switched over to linked-list style tracking of the next step, so the whole sequence number thing has gone away (at least for now).

functionary/core/models/workflow.py
scott-taubman: I don't think so, since the name here is just a friendly name mostly for human identification.  The `WorkflowStep` name has the restrictions because it has to be a legal name for use by the templating engine.

functionary/core/models/workflow_step.py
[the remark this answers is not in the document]
kvig: Yes, sorry. Missed the : vs =

functionary/core/models/workflow_run.py
kvig: This is still an issue but I think it's ok since you'd have to programatically trigger the workflow to try execute the next step.

kvig (approved):

xeonneon:
LGTM

PR #136, opened 2022-11-28
#114 - Scheduled Tasking frontend support

### LINKED ISSUE #114: Task scheduling
- [ ] Remove the scheduler app and existing routes, etc. (The task scheduling will be built into the core app)
- [ ] Evaluate scheduling libraries, both django specific and general python stuff (rolling our own is an option)
- [ ] Recurring date / time options at a minimum (Full on cron style would be a bonus)
- [ ] Models to store the schedule info
- [ ] Backend django management command that handles creating new tasks based on the scheduled jobs
- [ ] UI for managing schedules

xeonneon:
Closes #114 

## Introduced Changes
- Added scheduled task list, create, and update pages
- Dynamically load function parameters with htmx
- Dynamically validate crontab fields with htmx
- Added separate htmx urls
- Display message to user if they have no scheduled tasks they can make in their current environment
- Display form errors when creating or updating a `ScheduledTask`

## Testing
- Launch Functionary and view the `Scheduling` UI
- Create tasks and verify they run according to the crontab schedule
- Update tasks to verify their next run reflects those updates
- Try and submit invalid data to the forms and verify you receive an error

## Notes
- There are some helper functions in `core.utils.scheduling` that I think can be migrated over to the `ScheduledTask` model
- With my current implementation, I do not persist parameter arguments when a form gets rejected. The parameter values will reset to their defaults according to their widget type

### DIFF
--- a/functionary/ui/templates/base.html
+++ b/functionary/ui/templates/base.html
@@ -35,6 +35,7 @@
         <script src="{% static 'js/pretty-print-json-1.2.min.js' %}"></script>
         <!-- This script will run when the page loads and the menus are available -->
         <script src="{% static 'js/highlight-menu-item.js' %}" defer></script>
+        <script src="{% static 'js/htmx-1.8.4.min.js' %}"></script>
         {% unicorn_scripts %}
     </head>
     <body>
@@ -103,6 +104,12 @@
                             <a href="{% url 'ui:function-list' %}">
                                 <span class="icon has-text-info"><i class="fa fa-lg fa-cube"></i></span>
                                 &nbsp;Functions
+                            </a>
+                        </li>
+                        <li class="menu-item">
+                            <a href="{% url 'ui:schedule-list' %}">
+                                <span class="icon has-text-info"><i class="fa fa-lg fa-clock"></i></span>
+                                &nbsp;Schedules
                             </a>
                         </li>
                     </ul>
--- a/functionary/ui/templates/core/scheduled_task_create.html
+++ b/functionary/ui/templates/core/scheduled_task_create.html
@@ -1,0 +1,219 @@
+{% extends "base.html" %}
+{% load static %}
+{% block content %}
+<div class="block">
+    {% if form %}
+        <form id="scheduled-task-create-form"
+                method="post"
+                action="{% url 'ui:create-schedule' %}">
+            {% csrf_token %}
+            
+            <!-- Non-Field Errors display -->
+            <div class="columns is-centered">
+                <div class="column has-text-centered is-half">
+                    {% if form.non_field_errors %}
+                    <div class="notification is-warning has-text-black">
+                        <button class="delete" onclick="return this.parentNode.remove();"></button>
+                        Non-Field Errors:
+                        {{ form.non_field_errors }}
+                    </div>
+                    {% endif %}
+                </div>
+            </div>
+
+            <div class="box">
+
+                <!-- Name row/field -->
+                <div class="field">
+                    <div class="columns is-8">
+                        <div class="column is-one-third">
+                            <label for="{{ form.name.id_for_label }}">
+                                <b class="title is-size-4">{{ form.name.label }}</b> 
+                                <span class="is-small has-text-grey-light">{{ form.name.label_suffix }}</span>
+                                &nbsp;&nbsp;&nbsp;{{ form.name.help_text }}
+                            </label>
+                            <div class="column is-one-half p-0 is-italic subtitle is-size-6">
+                                <p>Give a descriptive name for your scheduled task.</p>
+                            </div>
+                            <div class="errorlist">
+                                {{ form.name.errors }}
+                            </div>
+                        </div>
+                        <div class="column">
+                            <div class="control">
+                                {{ form.name }}
+                            </div>
+                        </div>
+                    </div>
+                </div>
+
+                <hr/>
+
+                <!-- Description row/field -->
+                <div class="field">
+                    <div class="columns is-8">
+                        <div class="column is-one-third">
+                            <label for="{{ form.description.id_for_label }}">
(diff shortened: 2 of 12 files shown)


xeonneon:
Some additional Scheduled Task view features that could be added:
- Hide or un-hide the `Archived` scheduled tasks
- Sorting by scheduled task statuses (Active -> Error -> Paused -> Archived)
- Ability to immediately run a Scheduled Task

functionary/ui/templates/core/scheduled_task_list.html
```diff
@@ -0,0 +1,72 @@
...
+                        </td>
+                        <td>{{ scheduled_task.periodic_task.crontab }}</td>
+                        <td>{{ scheduled_task.status }}</td>
+                    </tr>
+                    {% endfor %}
+                </tbody>
+            </table>
+        </div>
+    </div>
+    <script src="{% static 'js/scheduled-task-list.js' %}" defer></script>
```
scott-taubman, line 71: This appears to be referencing a script that no longer exists.

functionary/ui/templates/base.html
```diff
@@ -86,6 +87,14 @@ <h1 class="title is-1">
                             </a>
                         </li>
                     </ul>
+                    <ul class="menu-list block">
```
scott-taubman, line 90: I vote to put this below "Functions" and to call it "Schedules" for the sake of consistency.

Also, the menu highlighting doesn't work because the javascript that drives that is keying off of `-list` being in the url name (i.e. schedule-list).  That might be a overly restrictive, but also having the url name be consistent with the others would probably be a good thing.

functionary/ui/templates/core/scheduled_task_create.html
```diff
@@ -0,0 +1,220 @@
+{% extends "base.html" %}
+{% load static %}
+{% load unicorn %}
```
scott-taubman, line 3: Left over unicorn stuff.

functionary/ui/templates/core/scheduled_task_update.html
```diff
@@ -0,0 +1,246 @@
+{% extends "base.html" %}
+{% load static %}
+{% load unicorn %}
```
scott-taubman, line 3: Another leftover

functionary/ui/templates/core/scheduled_task_create.html
```diff
@@ -0,0 +1,220 @@
...
+            </div>
+
+            <div class="box">
+
+                <!-- Name row/field -->
+                <div class="field">
+                    <div class="columns is-8">
+                        <div class="column is-one-third">
+                            <label for="{{ form.name.id_for_label }}">
+                                <b {% if form.name.errors %}class="errors"{% endif %} class="title">{{ form.name.label }}</b>
```
scott-taubman, line 32: The sizing of the title and subtitle end up being quite large and inconsistent with the other pages throughout the UI.  Can we shrink them a bit to make it feel more cohesive?  The same might be true of the text in the input boxes.

scott-taubman (commented):
Overall I think this looks good, worked as expected, and is organized pretty sensibly.  Most of my comments are about minor preferences regarding how some things look and some leftovers from the unicorn implementation.

Two things I'm slightly concerned about that I would be potentially okay addressing later:
* The scheduled_task_create.html and scheduled_task_update.html templates are overly similar for my tastes.  Figuring out how to reduce the duplication there would be nice.
* It appears that there is some re-use going on wrt the function parameter input rendering, but we should be conscious of that being a thing we're going to re-use in other places (i.e. workflows) and try to get it to as close as a drop-in reusable component as we can.

functionary/ui/views/scheduling.py
```diff
@@ -0,0 +1,285 @@
...
+    scheduled_task = ScheduledTask.objects.create(
+        name=data["name"],
+        environment=data["environment"],
+        description=data["description"],
+        function=data["function"],
+        parameters=task_params,
+        creator=request.user,
+    )
+
+    scheduled_task.periodic_task = create_periodic_task(data, scheduled_task)
```
scott-taubman, line 284: The `create_periodic_task` call also sets the value on the `scheduled_task` and saves it, so it doesn't seem like you have to actually assign it here, not that it hurts anything.

I might have structured the helpers a little different such that you just have a single call that creates the scheduled task and the periodic task for you, associating them properly etc. in a single transaction.  But that's not something I feel strongly enough about to say it needs to change here.

xeonneon: Fixed. I think in the future, I could probably just move the `create_periodic_task` to the `ScheduledTask` model as a helper function, so I don't need to pass the `ScheduledTask` object as reference

functionary/ui/templates/core/scheduled_task_create.html
xeonneon: I left the form labels as title/subtitles, but modified their sizes. Let me know if you think they look better. 

I chose to leave the input boxes the same, since I think the `is-small` modifier is way too small

functionary/ui/templates/core/scheduled_task_update.html
xeonneon: Fixed

functionary/ui/templates/core/scheduled_task_create.html
xeonneon: Fixed

functionary/ui/templates/base.html
xeonneon: Fixed. Also shortened the url names for scheduled tasks

functionary/ui/templates/core/scheduled_task_list.html
xeonneon: Fixed. Was a remnant from when I was going to do modals for the forms

Issue #142, opened 2022-12-01
Add pagination to the TaskList view


xeonneon:
When a user visits the `TaskList` view, they are currently presented with all of the Tasks. Add pagination to the view such that only batches of the most recent tasks are rendered at a time.

xeonneon:
Closed in #141

PR #145, opened 2022-12-02
#141 - Added pagination to TaskList, added ability to get task logs

### LINKED ISSUE #141: Display task logs in the UI
On the task details page it should be possible to view the task logs.  Some implementation notes:

* Don't load the actual log contents by default since they could potentially be very large.
* Could possibly have an accordion-style things similar to the way schema is listed on the functions, but instead of being loaded content that expands when you click it, the content isn't there until you click at which point htmx magic happens and the content is retrieved and loaded.

xeonneon:
Closes #141 and #142 

## Introduced Changes
- Added pagination navbar to the `TaskList` 
  - Created a pagination navbar partial so it can be re-used in other lists views
- Added button to `TaskDetail` to view the Task logs after the task has finished processing

## Testing
- View the `TaskList`, verify the pagination navbar is visible at the bottom
- View the `TastDetail`, wait for a task to finish executing, verify clicking the button to view logs returns the logs if they exist

### DIFF
--- a/functionary/ui/templates/partials/task_result.html
+++ b/functionary/ui/templates/partials/task_result.html
@@ -1,25 +1,7 @@
-<div id="result" class="block ml-4 is-12" hx-swap-oob="true">
-    {% if show_output_selector %}
-        <div class="tabs is-small is-toggle ml-4">
-            <ul hx-target="#result">
-                <li {% if output_format != "table" %}class="is-active"{% endif %}>
-                    <a hx-get="{% url 'ui:task-results' task.id %}?output=display_raw">Raw</a>
-                </li>
-                <li {% if output_format == "table" %}class="is-active"{% endif %}>
-                    <a hx-get="{% url 'ui:task-results' task.id %}?output=display_table">Table</a>
-                </li>
-            </ul>
-        </div>
-    {% endif %}
-    {% if not completed %}
-        <i id="result_indicator" class="fas fa-spinner fa-spin fa-2x"></i>
-    {% elif task.result is None %}
-        {% include "partials/output_none.html" %}
-    {% elif output_format == "json" %}
-        {% include "partials/output_json.html" %}
-    {% elif output_format == "table" %}
-        {% include "partials/output_table.html" %}
-    {% else %}
-        {% include "partials/output_string.html" %}
-    {% endif %}
-</div>
+{% if output_format == "log" %}
+    {% include "partials/task_log.html" %}
+{% elif task.result is None %}
+    {% include "partials/output_none.html" %}
+{% else %}
+    {% include "partials/output.html" with output_format=output_format %}
+{% endif %}
--- a/functionary/ui/urls.py
+++ b/functionary/ui/urls.py
@@ -85,6 +85,7 @@ urlpatterns = [
         (tasks.TaskDetailView.as_view()),
         name="task-detail",
     ),
+    path("task/<pk>/log", (tasks.get_task_log), name="task-log"),
     path(
         "task/<uuid:pk>/results",
         (tasks.TaskResultsView.as_view()),
--- a/functionary/ui/views/tasks.py
+++ b/functionary/ui/views/tasks.py
@@ -1,10 +1,19 @@
 import csv
 
-from django.core.exceptions import BadRequest
-from django.shortcuts import render
+from django.contrib.auth.decorators import login_required
+from django.core.exceptions import BadRequest, ValidationError
+from django.http import (
+    HttpRequest,
+    HttpResponse,
+    HttpResponseForbidden,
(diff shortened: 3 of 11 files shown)


functionary/ui/views/tasks.py
```diff
@@ -188,3 +199,30 @@ def get(self, request, *args, **kwargs):
...
+    if not request.user.has_perm(Permission.TASK_READ, env):
+        return HttpResponseForbidden()
+
+    if not (task_id := request.GET.get("task_id")):
+        return HttpResponse("Unknown task submitted.")
+
+    task = None
+    try:
+        # Use try/except in case of invalid task_id format
+        task = Task.objects.get(id=task_id, environment=env)
```
scott-taubman, line 217: There's a `get_object_or_404` helper in django.shortcuts or some such thing.  It handles this try except and response for you.  You may or may not still have to handle the `ValidationError` for malformed uuids.

functionary/ui/urls.py
```diff
@@ -166,6 +166,7 @@
         (scheduling.function_parameters),
         name="function-parameters",
     ),
+    path("task_log/", (tasks.get_task_log), name="task-log"),
```
scott-taubman, line 169: this path should probably be `/task/<pk>/log`

functionary/ui/views/tasks.py
```diff
@@ -188,3 +199,30 @@ def get(self, request, *args, **kwargs):
...
+
+
+@require_GET
+@login_required
+def get_task_log(request: HttpRequest) -> HttpResponse:
+    env = Environment.objects.get(id=request.session.get("environment_id"))
+    if not request.user.has_perm(Permission.TASK_READ, env):
+        return HttpResponseForbidden()
+
+    if not (task_id := request.GET.get("task_id")):
```
scott-taubman, line 211: The need for this check should go away if you change the path as suggesting in the comment on urls.py.

functionary/ui/views/tasks.py
```diff
@@ -188,3 +199,30 @@ def get(self, request, *args, **kwargs):
...
+    task = None
+    try:
+        # Use try/except in case of invalid task_id format
+        task = Task.objects.get(id=task_id, environment=env)
+    except ValidationError:
+        return HttpResponseNotFound("Unknown task submitted.")
+
+    task_log = TaskLog.objects.get(task=task)
+
+    if task_log.log == "":
```
scott-taubman, line 223: This whole check seems possibly superfluous.

scott-taubman (changes requested):

functionary/ui/templates/partials/task_result.html
```diff
@@ -15,11 +15,7 @@
         <i id="result_indicator" class="fas fa-spinner fa-spin fa-2x"></i>
     {% elif task.result is None %}
         {% include "partials/output_none.html" %}
```
kvig, line 17: No love for the None result? This could still have some logs to view.

kvig (changes requested):

functionary/ui/views/tasks.py
xeonneon: I confirmed that you do still need to handle the ValidationError for malformed UUID strings

functionary/ui/urls.py
xeonneon: Fixed. 

I tried moving the `tasks.get_task_log` to the existing `TaskResultsView`, but it already had a get method defined. Maybe it's worth just adding some additional logic to that view's get method to handle the `task/<pk>/log` endpoint? Not too sure if this is the correct approach, or if I should just make a new view

functionary/ui/views/tasks.py
xeonneon: Fixed

functionary/ui/templates/partials/task_result.html
xeonneon: That logic was already there, although I just changed it to be `task.result == ""`, since that logic would never be true. I believe `task.result` will always just be an empty string since its a `TextField`. Now if you view a task that did not produce any output (like the `long_running_task`), you should see the partial get rendered

xeonneon: Nevermind. This was because I was using old template images. Reverting back to the original logic

functionary/ui/urls.py
scott-taubman: Let's leave it as is for now.  I struggled a bit with this same thing in the workflows branch.  It seems inevitable that we're going to end up with a mix of view classes and one-off functions, but I'm sure there will be opportunities to consolidate or just generally more nicely organize things as we go along.

scott-taubman (approved):

Issue #140, opened 2022-12-01
Admin interface


scott-taubman:
Things that should be manageable via the admin interface:

* users
* tokens
* teams (just the main team model, not memberships)
* environments (same deal as teams)

Everything else that currently shows up there should go away.  For the items listed above, we should refine the process of managing that data in the admin interface where applicable, most specifically adding help text in places if possible.

PR #148, opened 2022-12-08
#140 - Refine the Admin interface.

kvig:
This PR cleans up the Admin interface.

It removes most of the entries in the Admin Interface, the header is made more consistent with the application and
Users/Teams/Environments are able to be seen from each other. Users may be assigned roles directly from the Team, Environment and User pages.


functionary/core/admin/team.py
```diff
@@ -0,0 +1,69 @@
...
+        # what we want.
+        if db_field.attname == "user_id":
+            field.widget = field.widget.widget
+        return field
+
+
+class TeamForm(forms.ModelForm):
+    """Custom form to show the environments associated with the team"""
+
+    environments = EnvironmentChoiceField(
```
scott-taubman, line 37: This ends up getting displayed as a multi-select box where your selection doesn't do anything.  Something like the list that you have for users, except it's read-only and just links over to the environments might be more appropriate.

functionary/core/admin/team.py
```diff
@@ -0,0 +1,69 @@
...
+    class Meta:
+        model = Team
+        fields = ["name"]
+
+    def __init__(self, *args, **kwargs):
+        super(TeamForm, self).__init__(*args, **kwargs)
+        # Since the Team doesn't have a direct reference to the
+        # Environments, add the queryset manually for this Team
+        if self.instance and "environments" in self.fields:
+            self.fields["environments"].queryset = Environment.objects.filter(
```
scott-taubman, line 51: You can get to the environments through the Team instance:

```python
self.instance.environments
```

functionary/ui/templates/admin/base_site.html
```diff
@@ -0,0 +1,81 @@
...
+                        </div>
+                    {% endif %}
+                </div>
+            </div>
+        </div>
+    </div>
+{% endblock branding %}
+{% block usertools %}
+{% endblock usertools %}
+{% block nav-sidebar %}
```
scott-taubman, line 73: There's a fly-out panel on the left side of the list and detail pages for models that broke somehow.  A javascript error shows up in the console and when you click the >> on the collapsed panel nothing happens.  I don't know what's causing that, but I'm guessing it's something in these template customizations.

scott-taubman (commented):
Made some comments about things I think could be cleaner, but for the most part looks good.

functionary/core/admin/team.py
kvig: Changed the list to readonly. I didn't see a widget that I could change it to in order to just display links. For a field on a model, that's handled by rendering a div on a template. This field is only defined on the form and makes things more challenging.

scott-taubman (commented):
I think this does the job for now.  We can further refine things as we use it and discover the pain points.

Issue #152, opened 2022-12-12
3rd party auth support (django-allauth possibly).


scott-taubman:
This is likely more of an epic than a single ticket.  The idea is that we want to support authentication through trusted 3rd parties; for example GitHub or GitLab (actual list or supported providers TBD).  Users would be able to authenticate with those providers rather than having a username & password that is specific to Functionary.

Some considerations, not all of which have to be dealt with in the initial implementation:

* The supported authentication mechanisms should ideally be able to be enabled or disabled individually, including the username / password auth.
* Individual auth providers need to also be configurable.  In the GitLab example, for instance, we need to allow for the location of the GitLab instance to be configured, since it may be some internal enterprise instance and not the public internet one.
* Whatever solution we come up with should either support a way to define a custom provider (any OAuth2 provider for instance) or make it possible to write a plugin to handle authentication from a custom provider.

scott-taubman:
Completed in #153

PR #153, opened 2022-12-19
Refs #152 - Add third party auth support

kvig:
This PR integrates `django-allauth` for supporting oauth2 and openid providers. This allows additional oauth providers to be configured for authentication in addition to the standard login/password.

When signing in with a third party auth, a User object will be created without access to any Functionary environments. When logging in without any environment access, a message will be displayed indicating this and to contact an admin to get access to environments instead of just showing a bunch of blank pages.

Clicking on the user name in the Functionary UI will display associated third party accounts and allow for disconnecting from the provider.

## Testing
To enable a third party auth via gitlab make the following modifications, adjusting `localhost` for your actual hostname.

Add the following line to the `settings/__init__.py`:
```
    "allauth.socialaccount.providers.gitlab",
```
Also, add the following to `settings/auth_.py`:
```python
SOCIALACCOUNT_PROVIDERS = {
    "gitlab": {
        "GITLAB_URL": "https://<gitlab url>",
    },
}
```

Go to *Gitlab* -> *User Settings* -> *Application*
Enter the following:
  - Name: "Functionary Auth API"
  - Redirect URI: "http://localhost:8000/accounts/gitlab/login/callback/"
  - Confidential: Checked
  - Scopes: read_user

Copy the `Application ID` and `Secret`
Log into the Django Admin
Add a *Site*: 
  - Set the `Domain Name` to the FQDN or the IP of the Django server
  - Set the `Display Name` to something meaningful.
  - 
Add a *Social Application*:
  - Select `GitLab` for the *Provider*
  - Set the *Name* to something meaningful
  - Set the *Client Id* to the `Application ID` from Gitlab
  - Set the *Secret Key* to the `Secret` from Gitlab
  - Select the *Site* created in the previous step and click the right pointing arrow to add it
Click `Save` to add the application.

You may need to set the following when launching Functionary if you get certificate errors when using a third party provider. If this is the case, make sure the certificate chain is correct, such as by setting the environment with:
```bash
REQUESTS_CA_BUNDLE = "/etc/ssl/certs/ca-certificates.crt"
```


scott-taubman:
Look at how GitLab organizes its login page as a possible source of inspiration wrt where the sign up and login with external providers stuff goes.

gitlab

Also worth noting is that it is pretty common for the login page to be basically it's own standalone page and not embedded within the normal page layout.  I mention that only because if styling the page so that it looks right with the functionary top banner, etc. was an issue, you can feel free to no worry about that.

functionary/ui/templates/core/function_detail.html
```diff
@@ -61,34 +61,36 @@ <h1 class="title is-1">
...
-                            </ul>
-                        </div>
-                    {% endif %}
-                    <input class="button {% if missing_variables %}is-warning{% else %}is-link{% endif %}"
-                           type="submit"
-                           value="Execute"/>
-                </div>
-            </form>
-        </div>
+        {% if form %}
```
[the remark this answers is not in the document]
kvig: Yes, if you didn't have permission to run the function you would get the parameter form I think without the execute button. That seemed unnecessary.

functionary/ui/templates/socialaccount/connections.html
```diff
@@ -0,0 +1,43 @@
+{% extends "base.html" %}
+{% block title %}
+    Account Connections
+{% endblock title %}
+{% block content %}
```
scott-taubman, line 5: Rather than having separate sections for removing and adding social account associations I wonder if it makes more sense to just list all of the available providers (the ones actually enabled, not the entire list supported by allauth), with the ones currently associated up top and the ones not associated on the bottom.  Then each one has an "Add" or "Remove" button depending on its current status.  

I'm including a super rough "mock-up" in the form of a spreadsheet.

image

scott-taubman:
Other than the comments made here, I'll note that we'll need to do some work on what the workflow is for admins to actually enable and configure providers.  We'll want to make that something that can be done via the django admin interface and smooth out as much of the process as we can, possibly abstracting away some of the allauth nitty-gritty.  That's something that I think should happen as a follow-on.

functionary/ui/templates/socialaccount/connections.html
kvig: Rewrote this page to be less confusing.

functionary/ui/templates/socialaccount/login.html
```diff
@@ -0,0 +1,19 @@
...
+    Sign In
+{% endblock title %}
+{% block content %}
+    <div class="content">
+        {% if process == "connect" %}
+            <h1>Connect {{ provider.name }}</h1>
+            <p>You are about to connect a new third party account from {{ provider.name }}.</p>
+        {% else %}
+            <h1>Sign In via {{ provider.name }}</h1>
+            <p>You are about to sign in using a third party account from {{ provider.name }}.</p>
```
scott-taubman, line 12: In the sign-in scenario this ends up looking a little jarring because it builds off of the base.html.  So you go from the clean login form over to a page with the Functionary banner for the redirect.  It's the kind of thing that we can smooth out later if we want, but I figured mention it.

functionary/ui/templatetags/social_helper.py
```diff
@@ -0,0 +1,12 @@
+from django import template
+
+register = template.Library()
+
+
+@register.simple_tag(takes_context=True)
+def find_account(context, provider_id):
```
scott-taubman, line 7: Let's try to remember docstrings on functions / methods.  Type hints are encouraged as well.

functionary/ui/admin/auth.py
```diff
@@ -0,0 +1,9 @@
+from allauth.account.adapter import DefaultAccountAdapter
+
+
+class FunctionaryAccountAdapter(DefaultAccountAdapter):
```
scott-taubman, line 4: Don't forget docstrings.

functionary/ui/templates/account/signup_closed.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <title>Functionary</title>
+        <meta http-equiv="refresh" content="0;URL={% url 'account_login' %}" />
+    </head>
+    <body>
+        <h1>Account signup is disabled.</h1>
```
scott-taubman, line 8: Similar question asked elsewhere:  How does one get to this page? Should it just go away?

functionary/functionary/urls.py
```diff
@@ -40,6 +40,7 @@
         name="redoc",
     ),
     path("ui/", include("ui.urls")),
+    path("accounts/", include("allauth.urls")),
```
scott-taubman, line 43: This feels like it belongs in the ui apps urls .  Is there a reason it has to be here?

functionary/ui/templates/socialaccount/login.html
kvig: Added a copy of Jace's modal to handle this.

There's some magic happening behind the scenes when you're not logged in to your gitlab account, like the token's not valid and it forces a different code path. The result is that you will always get prompted to confirm logging in via third party, vs just when you're not currently signed in to the thirdpary.

I tried getting around this by using the seemingly HTTP calls that the confirm uses(POST vs GET to the same URL), but I was getting CORS errors.

functionary/ui/templatetags/social_helper.py
```diff
@@ -0,0 +1,23 @@
...
+    templates context and returns the desired ProviderAccount.
+
+    Args:
+        context: the template's context, automatically passed in
+        provider_id: ID of the SocialAccount to find
+
+    Returns:
+        The matching ProviderAccount or None
+    """
+    if "form" in context:
```
xeonneon, line 19: Small nitpick: to keep this pythonic, this could just be `if context.get("form")`

functionary/functionary/settings/base/__init__.py
```diff
@@ -148,6 +153,7 @@
 AUTHENTICATION_BACKENDS = [
     "django.contrib.auth.backends.ModelBackend",
     "core.auth.backends.CoreBackend",
+    "allauth.account.auth_backends.AuthenticationBackend",
```
xeonneon, line 156: Could we instead just pre-configure several authentication providers, and then allow the admin user of the Functionary instance to toggle their enabled state? For example, we setup the basic configuration for Github, Gitlab, or Google out of the box. Then provide some admin interface to dynamically adjust the settings for those providers during runtime, or toggle whether or not the provider is enabled.

I am partly influenced by how [Pocketbase](https://pocketbase.io/) allow their admins to dynamically toggle and configure authentication providers during runtime, versus having to enable a provider prior to runtime: 
image
image

This isn't something that should be done in this PR, but just an idea to throw out there.

xeonneon (commented):
Everything looks good. Just some missing css files. 

Included some ideas that might be worth discussing. If possible, I'd be curious if we could  enable/disable/configure the authentication backends in a similar manner to Pocketbase

functionary/functionary/settings/base/__init__.py
scott-taubman: What you are describing here is effectively what the plan is and we've already got a follow-on ticket for the administration of the providers.  See #154.

functionary/ui/templates/socialaccount/login.html
scott-taubman: The behavior as you have it now seems good to me.

functionary/ui/templatetags/social_helper.py
scott-taubman: More specifically, something like:

```python
if form := context.get("form")
    for account in form.accounts:
       # carry on as before
```

functionary/ui/templates/account/login.html
```diff
@@ -0,0 +1,150 @@
...
+              sizes="16x16"
+              href="{% static 'favicon/favicon-16x16.png' %}"/>
+        <link rel="shortcut icon"
+              type="image/png"
+              href="{% static 'favicon/favicon.ico' %}"/>
+        <link rel="stylesheet"
+              href="{% static 'css/bulma-0.9.4.min.css' %}"
+              type="text/css"/>
+        <link rel="stylesheet" href="{% static 'css/brands.css' %}" />
+        <link rel="stylesheet" href="{% static 'css/fontawesome-6.2.1.min.css' %}" />
```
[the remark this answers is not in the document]
kvig: Good catch

functionary/ui/templates/account/login.html
```diff
@@ -0,0 +1,150 @@
...
+        {% block extracontent %}
+            {% get_providers as socialaccount_providers %}
+            <form>
+                {% csrf_token %}
+                <div class="container has-text-centered">
+                    <div class="columns">
+                        <div class="column is-6 container has-text-centered">
+                            <div class="card">
+                                <div class="card-content">
+                                    <span class="is-size-4">Or login with:</span>
```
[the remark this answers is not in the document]
kvig: Yes, hidden now

functionary/ui/templatetags/social_helper.py
kvig: fixed

PR #157, opened 2022-12-30
#156 - workflow utils

### LINKED ISSUE #156: WorkflowStep helpers
We need helper functions for adding, removing, and moving steps since performing any of these actions requires manipulating the "next" pointer on other steps in the workflow.

scott-taubman:
Closes #156

This PR adds a few helper utilities for manipulating workflows.  It also includes some other needed backend tweaks related to workflows.

* Adds the new CRUD permissions
* Adds a helper on the model for getting an ordered list of the workflow steps
* Removes schema from the workflow (for now) in favor of a WorkflowParameter model.

## Testing Instructions

Unit tests have been included to test and demonstrate the functionality of the 3 new utility functions as well as the new model helper.


functionary/core/utils/workflow.py
```diff
@@ -0,0 +1,103 @@
...
+
+    with transaction.atomic():
+        if before:
+            before.next = step.next
+            before.save()
+
+        step.delete()
+
+
+def move_step(step: WorkflowStep, next: WorkflowStep | None = None) -> None:
```
kvig, line 75: Can this function get a more descriptive name like `move_step_before` since the parameters are both the same type?

functionary/core/utils/workflow.py
```diff
@@ -0,0 +1,103 @@
...
+    Returns:
+        The created WorkflowStep
+
+    Raises:
+        ValueError: workflow and next.workflow do not match
+    """
+    if next is not None and workflow != next.workflow:
+        raise ValueError("Provided next step is not part of provided workflow")
+
+    previous_step = WorkflowStep.objects.filter(workflow=workflow, next=next).first()
```
kvig, line 38: Nit: can you make these variable names consistent across the functions, this one is `previous` and the other two are `before` and `before_step` steps.

functionary/core/models/workflow_parameter.py
```diff
@@ -0,0 +1,49 @@
+from django.core.validators import RegexValidator
+from django.db import models
+
+PARAMETER_TYPES = [
```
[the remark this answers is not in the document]
scott-taubman: Yeah, I originally had a TODO here to find a better home for this where it would make sense to reuse it between the two.  I briefly had it in the Function model, which feels like a better home, except that it isn't actually used by Function, so that felt weird.  If a `FunctionParameter` model were a thing, that would definitely be the right spot.  Maybe that should be a thing...That's a whole 'nother conversation though.

functionary/core/utils/workflow.py
scott-taubman: Changed to `before_step` to match the others.

functionary/core/utils/workflow.py
scott-taubman: I feel like the parameter names and the docstring make it mostly clear enough so I'm inclined to keep the name as is unless you really feel flip the table over strong on this one.

kvig (approved):

PR #160, opened 2023-01-05
Move ScheduledTaskForm to its own file

scott-taubman:
This PR does two things:

* Moves ScheduledTaskForm into its own file
* Makes it possible to override the field type mapping in TaskParameterForm, in preparation for some work to come related to Workflows.

## Testing Instructions
* Verify that it is still possible to create / edit scheduled tasks.
* NOTE: There are some bugs related to scheduled tasks (json fields do not work, warning message displays twice, maybe others), but there should be no new bugs introduced by this change.

### DIFF
--- a/functionary/ui/forms/__init__.py
+++ b/functionary/ui/forms/__init__.py
@@ -1,0 +1,2 @@
+from .scheduled_task import ScheduledTaskForm  # noqa
+from .tasks import TaskParameterForm  # noqa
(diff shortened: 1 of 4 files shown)


xeonneon:
The warning message getting displayed twice is due to the additional message rendering in `auth_messages.html`. Maybe it makes sense to rename the `auth_messages.html` to `messages.html` and use that as the global message display partial?

PR #163, opened 2023-01-06
Add scheduler logging and vscode debug task

scott-taubman:
This adds a vscode debug task and enables logging for the scheduler.

## Test Instructions
* Start everything as usual.
* View the logs on the scheduler container.  You should see messages from beat about what it's doing.
* In vscode there's a new "run_scheduler" debug task.  You can launch that to start the scheduler locally in debug mode.


functionary/functionary/settings/base/logging_.py
```diff
@@ -37,6 +37,7 @@
         "docker": {"propagate": False},
         "kombu": {"handlers": ["console"], "propagate": False},
         "urllib3": {"propagate": False},
+        "celery.beat": {"propogate": False},
```
kvig, line 40: propagate is spelled wrong

functionary/core/management/commands/run_scheduler.py
```diff
@@ -7,6 +8,10 @@
 from core.celery import app
 from core.utils.messaging import initialize_messaging, wait_for_connection
 
+LOG_LEVEL = getenv("LOG_LEVEL", "INFO")
+logger = logging.getLogger("celery.beat")
+logger.setLevel(getattr(logging, LOG_LEVEL))
```
kvig, line 13: Is it possible to get this from the django.config?

functionary/.vscode/launch.json
```diff
@@ -47,6 +47,17 @@
...
+    {
+      "name": "run_scheduler",
+      "type": "python",
+      "request": "launch",
+      "preLaunchTask": "scheduler_stop_container",
+      "program": "${workspaceFolder}/manage.py",
+      "args": ["run_scheduler"],
+      "envFile": "${workspaceFolder}/.vscode/.debug.env",
+      "django": true,
+      "justMyCode": false
```
kvig, line 60: Did you mean to default this to false?

scott-taubman: Yes, this was on purpose.  The scheduler is pretty much all celery code, we just wrap it with a django management command to launch it.  So if this is set to True, there's not much to meaningful step through in the even of doing some debugging.

Issue #167, opened 2023-01-13
Refactor ScheduledTasks to align with new View directory structure


xeonneon:
Refactor the ScheduledTasks views to follow the new View directory structure, as well as try to use some more Django specific tools where they make sense.

PR #168, opened 2023-01-13
#161 - Fixed serialization issues during validation

### LINKED ISSUE #161: ScheduledTasks do not support json parameters
Attempting to schedule a task with json parameters fails because the json is always treated as invalid.

xeonneon:
Closes #161 

## Introduced Changes
- Added new util to handle serialization of python dictionaries to JSON strings
  - This utility is used to create a new dictionary of a Task or ScheduledTask parameters, which is what the JSON Schema Validator will use to validate the parameters conform to the function's schema
- Removed the `_prepare_initial_value` function in the Tasks.py since it was really only ever looking for a JSON field to run `loads`. Now, in the `__init__()` of the `TaskParameterForm`, the form just uses the value from initial.
- Added some missing type hints
- Fixed some linting complaints
- Fixed missing ScheduledTask url for form submission
- Removed `obj = *.objects.create()` in favor of `obj = Object() -> obj.clean()  -> obj.save()`
- Removed extra message display in ScheduledTask list in favor of the message display used by the auth templates

## Testing
- Run some functions with valid and invalid parameters. Ensure you are getting appropriate errors.
- Create and Update some ScheduledTasks. Verify the Schedules are validating the input

## Notes
- The serialization function for the parameters is limited to the current nesting amount as shown in the `output_json` example schema. If we ever want to include more nesting in the schema, the serialization function will need to be updated to a more dynamic solution

### DIFF
--- a/functionary/core/utils/serialization.py
+++ b/functionary/core/utils/serialization.py
@@ -1,0 +1,90 @@
+from copy import deepcopy
+from json import JSONDecodeError, dumps, loads
+from typing import Union
+
+# Place helper methods for serializing data that is used by the models in here
+# Don't import any models to prevent cyclic module dependencies
+
+
+def serialize_parameters(parameters: dict, schema: dict) -> Union[dict, None]:
+    """JSON stringify parameters which are supposed to be JSON formatted strings
+
+    Iterate throw the given schema to find any function arguments that
+    could be JSON formatted strings. If so, dump the value of that argument
+    into a JSON string, and overwrite that value in a deep copy of the
+    given parameters dictionary.
+    Perform the same logic if the function argument can be multiple different
+    types, as indicated by 'anyOf'.
+    Returns a deep copy of the parameters that contains all of the serialized
+    JSON fields.
+
+    NOTE: This function does not handle nested schemas past the function
+    argument->anyOf[{}, {}, ...]. If the function schemas need to include more
+    nesting, this will need to be reworked.
+
+    Args:
+        parameters: A dictionary that contains the parameters for the function.
+            Example format looks like {'a': 1, 'b': 2}
+        schema: A dictionary that represents the JSON schema for the function
+
+    Returns:
+        parameters: A new dictionary that contains all of the serialized
+            function arguments.
+    """
+    if not parameters:
+        return {}
+
+    parameters_copy = deepcopy(parameters)
+    try:
+        for arg, param_type in schema["properties"].items():
+            if union_type_param := param_type.get("anyOf"):
+                serialize_union_parameters(arg, union_type_param, parameters_copy)
+            else:
+                if is_json_field(param_type):
+                    parameters_copy[arg] = dumps(parameters_copy[arg])
+                    # Validate the serialized JSON string is valid JSON
+                    _ = loads(parameters_copy[arg])
+        return parameters_copy
+    except JSONDecodeError as err:
+        raise err
+
+
+def serialize_union_parameters(
+    arg: str, union_type_param: dict, parameters: dict
+) -> None:
+    """Mutate given parameters dictionary to json stringify the parameters"""
+    for param_type in union_type_param:
+        if is_json_field(param_type):
+            parameters[arg] = dumps(parameters[arg])
+            _ = loads(parameters[arg])
+
+
+def json_stringify_parameters(parameters: dict) -> str:
+    """Turns the values of the dictionary keys into JSON strings
+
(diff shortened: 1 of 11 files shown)


functionary/core/utils/serialization.py
```diff
@@ -0,0 +1,66 @@
...
+    parameters_copy = deepcopy(parameters)
+    try:
+        for arg, param_type in schema["properties"].items():
+            if union_type_param := param_type.get("anyOf"):
+                serialize_union_parameters(arg, union_type_param, parameters_copy)
+            else:
+                if is_json_field(param_type):
+                    parameters_copy[arg] = dumps(parameters_copy[arg])
+                    # Validate the serialized JSON string is valid JSON
+                    _ = loads(parameters_copy[arg])
```
scott-taubman, line 49: This step shouldn't be necessary since you generated the JSON with dumps.  If `parameters_copy[arg]` is not actually json serializable the `dumps` would throw an exception.

functionary/core/utils/serialization.py
```diff
@@ -0,0 +1,66 @@
...
+        return parameters_copy
+    except JSONDecodeError as err:
+        raise err
+
+
+def serialize_union_parameters(
```
scott-taubman, line 55: By convention, `serializer_union_parameters` and `is_json_field` should start with an `_`, since they seem like they're intended as internal helpers here and not expected to be imported and used elsewhere.

functionary/ui/forms/tasks.py
```diff
@@ -122,7 +111,7 @@ def __init__(self, function, data=None, initial=None, prefix="task-parameter"):
             kwargs = {
                 "label": value["title"],
                 "label_suffix": param_type,
-                "initial": _prepare_initial_value(param_type, initial_value),
```
scott-taubman, line 125: I know there was some back and forth, but we landed on this needing to stay.  The thing that needs to change is that `_transform_initial_mapping` should no longer just point to `json.loads` as the transformation function, but instead some other helper that calls `json.loads` only if necessary (the value is actually a string).

As it is now, if someone provides a default value for a json parameter, the value gets populated in the form as an escaped string, like:

`"{\"param1\": \"value1\", \"param2\": \"value2\"}"`

The whole `_preapre_initial_value` thing is there to make sure it instead is populated as:

`{"param1": "value1", "param2": "value2"}`

functionary/ui/views/scheduling.py
```diff
@@ -269,14 +275,18 @@ def _create_scheduled_task(
 ):
     """Helper function for creating scheduled task"""
     with transaction.atomic():
-        scheduled_task = ScheduledTask.objects.create(
+        scheduled_task = ScheduledTask(
```
scott-taubman, line 278: This is a case where using `objects.create()` or skipping the `clean()` is okay.  You are basically saying as much in your comment below because you are expecting the data to come in pre-validated.  As is, you're validating the data twice without any changes having happened in between.

The validation isn't especially expensive in this case and I suppose it does protect you from a scenario where someone messes up and passes non-validated data into this function, but given that it's an internal helper, I think trusting it and just noting in the docstring that the data is assumed to have be pre-validated would be acceptable.

functionary/ui/urls.py
```diff
@@ -140,6 +140,11 @@
         (scheduling.ScheduledTaskCreateView.as_view()),
         name="new-schedule",
     ),
+    path(
```
scott-taubman, line 143: I'm assuming this got accidentally removed in some other PR or merge conflict process?

functionary/core/utils/serialization.py
xeonneon: Fixed

functionary/core/utils/serialization.py
xeonneon: Yes that is correct. Fixed

functionary/ui/forms/tasks.py
xeonneon: Implemented the helper function

functionary/ui/views/scheduling.py
xeonneon: Ok, makes sense. I wasn't sure if we wanted to just switch to the object, clean, save flow for everything, so I had included that comment. Fixed

functionary/ui/urls.py
xeonneon: Yea, I may have accidentally removed it when rebasing the team/environment management stuff. Although, these urls are going to be refactored in #167

scott-taubman (approved):

Issue #143, opened 2022-12-01
Workflow Creation Frontend


scott-taubman:
- [x] Basic workflow creation
- [x] Add / edit / remove workflow parameters
- [x] Add / remove workflow steps
- [x] Re-order workflow steps

PR #170, opened 2023-01-17
#143 - Workflow Steps UI

scott-taubman:
Continues #143 

This PR adds the ability to create, delete, and edit workflow steps.  Some related things that happen in this PR:

* Created TaskParameterTemplateForm, a variant of TaskParameterForm that accepts template variables as input in addition to the field type.
* Updated the "function_parameters/" endpoint to take an optional "allow_template_variables" parameter, which decides which variant of the TaskParameterForm gets used.
* Because "function_parameters/" is no longer ScheduledTask specific, it has been relocated to a more appropriate, general location.
* The bit in "function_parameters" that derives the existing parameters so that it can populate the form fields for existing scheduled tasks has been removed.  Instead, the template for updating scheduled tasks just includes the parameter template and no longer does the lazy loading thing.

NOTE: As with much of this Workflow stuff, the UI is purely aiming to be functional.  The goal is to demonstrate the functionality and prove out the general behavior.

## Test Instructions
* Create a workflow and then go to the workflow details page.
* Below the parameters is a new section for steps.
* Play around with adding, editing, and deleting steps.
* The + buttons on a step row inserts a new step below the current one.
* Verify that typed parameters accept only valid input of that type, or a template variable.  For example, an integer parameter should behave as follows:
  * 123 (valid)
  * 12.3 (invalid)
  * somestring (invalid)
  * {{somestring}} (valid)

### DIFF
--- a/functionary/ui/forms/tasks.py
+++ b/functionary/ui/forms/tasks.py
@@ -48,7 +48,11 @@ _field_mapping = {
 
 def _transform_json(value: Union[str, dict]) -> Union[str, dict]:
     if type(value) is str:
-        return json.loads(value)
+        try:
+            return json.loads(value)
+        except json.JSONDecodeError:
+            pass
+
     return value
 
 
@@ -194,7 +198,7 @@ class TaskParameterTemplateForm(TaskParameterForm):
         self,
         function: "Function",
         data: Optional["QueryDict"] = None,
-        initial: str | None = None,
+        initial: Optional[str] = None,
     ):
         """TaskParameterTemplateForm init
 
--- a/functionary/ui/forms/tasks.py
+++ b/functionary/ui/forms/tasks.py
@@ -1,5 +1,6 @@
 import json
-from typing import Tuple, Type, Union
+import re
+from typing import TYPE_CHECKING, Optional, Tuple, Type, Union
 
 from django.forms import (
     BooleanField,
@@ -12,10 +13,14 @@ from django.forms import (
     IntegerField,
     JSONField,
     Textarea,
+    TextInput,
 )
 from django.forms.widgets import DateInput, DateTimeInput, Widget
 
-from core.models import Function
+if TYPE_CHECKING:
+    from django.http import QueryDict
+
+    from core.models import Function
 
 
 class HTMLDateInput(DateInput):
@@ -112,9 +117,9 @@ class TaskParameterForm(Form):
 
     def __init__(
         self,
-        function: Function,
-        data: dict = None,
-        initial: dict = None,
+        function: "Function",
+        data: dict | None = None,
+        initial: dict | None = None,
         prefix: str = "task-parameter",
     ):
         super().__init__(data=data, prefix=prefix)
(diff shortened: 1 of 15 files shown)


functionary/ui/forms/tasks.py
```diff
@@ -160,3 +165,76 @@ def _get_field_info(
...
+        Args:
+            function: The Function that will be executed when the WorkflowStep runs
+            data: Form submission data
+            initial: Initial values to populate the form fields with on render
+        """
+        initial_data = (
+            json.loads(self._stringify_template_variables(initial)) if initial else {}
+        )
+
+        super().__init__(function=function, data=data, initial=initial_data)
```
kvig, line 210: I got the following error when trying to edit a step that I set the `load_json()` parameter to `{{param1}}`:
`json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)`

kvig (changes requested):
It would be nice to be able to create a new step 1 without deleting the existing one and remaking it, but it's ok for that or dragging rearrange to be implemented in a follow-on ticket

functionary/ui/forms/tasks.py
```diff
@@ -160,3 +165,76 @@ def _get_field_info(
...
+        template"""
+        json_data = json.dumps(parameters)
+
+        return re.sub(r'"{{([\w\.]*)}}"', r"{{\1}}", json_data)
+
+    def __init__(
+        self,
+        function: "Function",
+        data: Optional["QueryDict"] = None,
+        initial: str | None = None,
```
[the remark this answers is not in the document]
scott-taubman: I only recently learned that the `|` form was a thing and prefer it since it doesn't require importing stuff from `typing`.  Unfortunately, you can't use that form with types that are only import for typing purposes (the stuff under `if TYPE_CHECKING`).  Thus the use of `Optional["QueryDict"]`.  Since I'm importing `Optional` anyway in this case, I switched to using that in both cases to keep the definitions consistent here.

functionary/ui/forms/workflow_step.py
```diff
@@ -0,0 +1,63 @@
...
+    template_name = "forms/workflows/step_edit.html"
+
+    class Meta:
+        model = WorkflowStep
+        fields = ["workflow", "name", "function", "next"]
+        widgets = {"workflow": forms.HiddenInput(), "next": forms.HiddenInput()}
+
+    def __init__(
+        self,
+        environment: Optional["Environment"] | "UUID" | str = None,
```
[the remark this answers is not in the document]
scott-taubman: It ends up the same, but I agree that's a more straightforward way to define it.  Updated.

functionary/ui/forms/tasks.py
scott-taubman: I fixed this in a way that I think I can live with for now.  Basically, if the json loads fails for the initial values, I just have it return the original string value.  The result is a little bit weird because when you go to edit a step with a json parameter containing a template variable, the variable now shows up as quoted.  

image

The behavior ends up being correct, whether you leave the quotes in or take them out, so I think I can live with that for now.

I also added a unit test for the scenario that was failing, which now passes with the change.

kvig: That works for me.

kvig (approved):

PR #175, opened 2023-01-20
#167 - ScheduledTask now use class based views

### LINKED ISSUE #167: Refactor ScheduledTasks to align with new View directory structure

xeonneon:
Closes #167 

## Introduced Changes
- Restructured original ScheduledTask implementation to use the new Views directory structure layout
- ScheduledTask views are now class based views
- Added a new detail view for ScheduledTasks which includes a paginated list of tasks spawned by the ScheduledTasks

## Testing
- Create and update Scheduled Tasks just like before


functionary/ui/views/schedules/update.py
```diff
@@ -0,0 +1,85 @@
...
+        environment = get_object_or_404(
+            Environment, id=self.request.session.get("environment_id")
+        )
+        kwargs["environment"] = environment
+        return kwargs
+
+    def post(self, request: HttpRequest, **kwargs) -> HttpResponse:
+        scheduled_task: ScheduledTask = self.get_object()
+        data: QueryDict = request.POST.copy()
+        data["environment"] = scheduled_task.environment
```
scott-taubman, line 51: Two things here:

* You could create a separate `ScheduledTaskUpdateForm` that excludes the environment field, since you don't want that to change.  Then you wouldn't worry about explicitly setting it here.  See the `WorkflowStepCreateForm` and `WorkflowStepUpdateForm` as an example, which does what I'm describing with the "workflow" field, since that should never be updated after initial creation.

* I'm kind of of the opinion that the function shouldn't change on a scheduled task after initial creation either.  As soon as you change the function it is for all intents an purposes an entirely new scheduled task, just weirdly associated with the tasking history of some other scheduled task.  If we were tracking the revision history of the scheduled tasks I could see maybe allowing it, but for now I'd vote to not.

functionary/ui/views/schedules/update.py
```diff
@@ -0,0 +1,85 @@
...
+from ui.views.view_base import PermissionedFormUpdateView
+
+
+class ScheduledTaskUpdateView(PermissionedFormUpdateView):
+    model = ScheduledTask
+    form_class = ScheduledTaskForm
+    template_name = "forms/schedules/scheduling_create_or_update.html"
+
+    def get_success_url(self) -> str:
+        return reverse("ui:schedule-list")
```
scott-taubman, line 18: Now that you have the detail view, should that be where the success_url points?

functionary/ui/views/schedules/utils.py
```diff
@@ -51,6 +51,19 @@ def get_or_create_crontab_schedule(
     return crontab_schedule
 
 
+def get_crontab_schedule(schedule_form_data: dict) -> CrontabSchedule:
```
scott-taubman, line 54: This function belongs here, but the rest of these were more appropriate living in `core/utils/scheduling.py`

The distinction is that this function deals with UI specific stuff: it's parsing the form data and using it call the general `get_or_create_crontab_schedule`.

The rest of the functions in this file have applicability outside of the UI and could conceivably called in other contexts (as part of API request handline for instance).

Basically, leave the old utils file as it was and create this as a new file here with your new function.

scott-taubman (changes requested):

functionary/ui/views/view_base.py
```diff
@@ -54,10 +54,10 @@ def test_func(self):
 class PermissionedFormCreateView(LoginRequiredMixin, UserPassesTestMixin, CreateView):
     def test_func(self):
         env = self.request.session.get("environment_id")
-        return self.request.user.has_perm(Permission.ENVIRONMENT_READ, env)
+        return self.request.user.has_perm(Permission.ENVIRONMENT_CREATE, env)
```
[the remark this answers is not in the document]
xeonneon: Fixed

functionary/ui/views/schedules/detail.py
```diff
@@ -0,0 +1,62 @@
...
+            "-created_at"
+        )
+        paginator = Paginator(history, PAGINATION_AMOUNT)
+        page_obj = paginator.get_page(self.request.GET.get("page"))
+        context["paginator"] = paginator
+        context["page_obj"] = page_obj
+        return context
+
+
+# TODO: Move this endpoint somewhere else if it is going to be used by other views
```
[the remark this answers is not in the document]
xeonneon: Fixed

functionary/ui/views/schedules/detail.py
```diff
@@ -0,0 +1,62 @@
...
+    template_name = "core/scheduling_detail.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        scheduledtask: ScheduledTask = self.get_object()
+        history = Task.objects.filter(scheduled_task=scheduledtask).order_by(
+            "-created_at"
+        )
+        paginator = Paginator(history, PAGINATION_AMOUNT)
+        page_obj = paginator.get_page(self.request.GET.get("page"))
```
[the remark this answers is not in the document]
xeonneon: Changed it so that it just shows the most recent 10 tasks and removed the Pagination object from the page

functionary/ui/templates/forms/schedules/scheduling_create_or_update.html
```diff
@@ -49,7 +53,8 @@
 
                 <hr/>
 
-                <!-- Status row/field -->
+                {% comment %} Should only render on update page {% endcomment %}
```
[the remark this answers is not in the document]
xeonneon: Fixed

functionary/ui/urls.py
```diff
@@ -118,53 +118,33 @@
...
-        (scheduling.ScheduledTaskUpdateView.as_view()),
-        name="schedule-detail",
-    ),
-    path(
-        "schedule_list/",
-        (scheduling.ScheduledTaskListView.as_view()),
-        name="schedule-list",
+        "function_parameters/",
+        (schedules.function_parameters),
+        name="function-parameters",
```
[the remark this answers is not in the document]
xeonneon: Fixed

functionary/ui/templates/forms/schedules/scheduling_create_or_update.html
```diff
@@ -3,10 +3,14 @@
 {% block content %}
 <div class="block">
     {% if form %}
-        <form id="scheduled-task-create-form"
-                method="post"
-                action="{% url 'ui:update-schedule' scheduledtask.id %}">
-            {% csrf_token %}
+        
+        {% if update %}
```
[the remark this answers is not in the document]
xeonneon: Fixed

functionary/ui/forms/scheduled_task.py
```diff
@@ -36,59 +36,83 @@ class ScheduledTaskForm(ModelForm):
...
         self._setup_field_classes()
 
-    def clean(self):
-        cleaned_data = super().clean()
-        available_functions = Function.objects.filter(
-            package__environment=cleaned_data["environment"]
-        )
-        if cleaned_data["function"] not in available_functions:
+    def clean_function(self):
+        """The function field needs to be declared AFTER the environment field"""
```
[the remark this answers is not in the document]
xeonneon: I had added the clean method for the function just for the assurance that the user was only scheduling a function that they had access to. I removed it since it is redundant

functionary/ui/views/schedules/update.py
```diff
@@ -0,0 +1,85 @@
...
+        initial["scheduled_minute"] = crontab.minute
+        initial["scheduled_hour"] = crontab.hour
+        initial["scheduled_day_of_week"] = crontab.day_of_week
+        initial["scheduled_day_of_month"] = crontab.day_of_month
+        initial["scheduled_month_of_year"] = crontab.month_of_year
+        return initial
+
+    def get_context_data(self, **kwargs) -> dict:
+        context = super().get_context_data(**kwargs)
+        scheduled_task: ScheduledTask = self.get_object()
```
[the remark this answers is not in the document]
xeonneon: Fixed

functionary/ui/views/schedules/update.py
xeonneon: Since the update view already has context of the object its editing, I just stamp in the environment and the function from the current object, which basically makes those fields of the ScheduledTask immutable. I disabled the function field so that the user knows they shouldn't be editing the function.

functionary/ui/views/schedules/update.py
xeonneon: Fixed

functionary/ui/views/schedules/utils.py
xeonneon: Fixed

functionary/ui/forms/scheduled_task.py
```diff
@@ -36,59 +36,70 @@ class ScheduledTaskForm(ModelForm):
...
             "description",
             "status",
             "function",
             "parameters",
-            "environment",
         ]
 
     def __init__(self, *args, **kwargs):
-        env: Environment = kwargs.pop("env")
+        environment = kwargs.pop("environment", None)
```
scott-taubman, line 56: Somethings I probably should have noticed before, so apologies:

It'd probably be better to just list environment in the signature of the `__init__` and then not pass it along when you call super:

```python
def __init__(self, environment: Environment, *args, **kwargs):
    # Now this call doesn't include environment because it's not bundled in with **kargs
    super().__init__(*args, **kwargs)
```

As for `is_create`, I don't think that should be necessary as a parameter.  That info is implicit in the presence or absence of the `instance` parameter.  When doing an update, that would be set to the instance you are updating.  If that's not set, then you are creating a new instance.

functionary/ui/views/schedules/detail.py
```diff
@@ -0,0 +1,19 @@
+from core.models import Task
+from core.models.scheduled_task import ScheduledTask
+from ui.views.view_base import PermissionedEnvironmentDetailView
+
+PAGINATION_AMOUNT = 15
```
scott-taubman, line 5: This can presumably be removed now.

functionary/ui/views/view_base.py
```diff
@@ -60,4 +60,4 @@ def test_func(self):
 class PermissionedFormUpdateView(LoginRequiredMixin, UserPassesTestMixin, UpdateView):
     def test_func(self):
         env = self.request.session.get("environment_id")
-        return self.request.user.has_perm(Permission.ENVIRONMENT_READ, env)
+        return self.request.user.has_perm(Permission.ENVIRONMENT_CREATE, env)
```
scott-taubman, line 63: This should be changed back for now as well.

functionary/ui/forms/scheduled_task.py
xeonneon: Fixed

functionary/ui/views/schedules/detail.py
xeonneon: Fixed

functionary/ui/views/view_base.py
xeonneon: Fixed

functionary/ui/forms/scheduled_task.py
```diff
@@ -52,11 +52,9 @@ class Meta:
             "parameters",
         ]
 
-    def __init__(self, *args, **kwargs):
-        environment = kwargs.pop("environment", None)
-        is_create = kwargs.pop("is_create", False)
+    def __init__(self, environment: Environment = None, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self._setup_field_choices(is_create)
+        self._setup_field_choices(kwargs.get("instance"))
```
scott-taubman, line 57: You're passing in an instance (or None) to `is_update` which is listed as a bool.  Probably just pass in `kwargs.get("instance") is not None`, and then everything lines up more cleanly.

xeonneon: Fixed

scott-taubman (approved):

PR #176, opened 2023-01-24
#127 - Add active field to functions #127

### LINKED ISSUE #127: Functions that are removed from a package.yaml are not properly handled.
If a package is published with a previously present function removed, the function is still present and remains taskable via the UI and API.  Instead, what should happen is that the function should be deactivated such that it no longer shows up in the UI or API list views and it cannot be tasked.

lights7412:
Closes #127
The main objective of this PR is to add the active field to the function model. The functions listed in the package should reflect only the ones listed in the package.yaml.

**Testing Instructions**
A unit test has been added to test the deactivation feature in the function.
Verify that the only functions that appear in the package UI are the ones in the yaml.
In addition, the drop down list to schedule should only have the active functions as options.
The inactive functions should cause scheduled tasks related to that function to be paused.
Verify the inactive functions detail page do not have an execute button and should have a banner saying its deactivated
Removing a function from the yaml should result in any scheduled tasks associated to be paused, although that function details should still be viewable.

### DIFF
--- a/functionary/builder/utils.py
+++ b/functionary/builder/utils.py
@@ -197,6 +197,7 @@ def build_package(build_id: UUID):
         db_functions = _create_functions_from_definition(
             package_definition.get("functions"), package
         )
+        _deactivate_functions(package_definition.get("functions"), package)
         _extract_package_contents(package_contents, workdir)
         _load_dockerfile_template(dockerfile, workdir)
 
@@ -305,9 +306,25 @@ def _create_functions_from_definition(definitions: list[dict], package: Package)
         function_obj.schema = _generate_function_schema(
             name, function_def.get("parameters")
         )
+        function_obj.active = True
         db_functions.append(function_obj)
 
     return db_functions
+
+
+def _deactivate_functions(definitions, package: Package):
+    """Grabs all functions from DB, deactivates any that aren't in the definition"""
+    function_names = []
+
+    for function_def in definitions:
+        function_names.append(function_def.get("name"))
+
+    removed_functions = Function.objects.exclude(
+        package=package, name__in=function_names
+    )
+
+    for function in removed_functions:
+        function.deactivate()
 
 
 def _generate_function_schema(name: str, parameters) -> str:
--- a/functionary/core/models/function.py
+++ b/functionary/core/models/function.py
@@ -28,6 +28,7 @@ class Function(models.Model):
         variables: list of variable names to set before execution
         return_type: the type of the object being returned
         schema: the function's OpenAPI definition
+        active: whether the function is currently activated
     """
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
@@ -43,6 +44,7 @@ class Function(models.Model):
     variables = models.JSONField(default=list, validators=[list_of_strings])
     return_type = models.CharField(max_length=64, null=True)
     schema = models.JSONField()
+    active = models.BooleanField(default=True)
 
     class Meta:
         constraints = [
@@ -69,6 +71,14 @@ class Function(models.Model):
 
     def clean(self):
         self._clean_environment()
+
+    def deactivate(self):
+        """Deactivate the function and pause any associated scheduled tasks"""
+        self.active = False
+        self.save()
+
+        for scheduled_task in self.scheduled_tasks.all():
+            scheduled_task.pause()
 
     @property
     def render_name(self) -> str:
--- a/functionary/core/models/task.py
+++ b/functionary/core/models/task.py
@@ -98,10 +98,16 @@ class Task(ModelSaveHookMixin, models.Model):
         except JSONDecodeError as err:
             raise ValidationError(err.msg)
 
+    def _clean_function(self):
+        """Validate function is active for newly created tasks"""
+        if self._state.adding and self.function.active is False:
+            raise ValidationError("This function is not active")
+
     def clean(self):
         """Model instance validation and attribute cleanup"""
         self._clean_environment()
         self._clean_parameters()
+        self._clean_function()
 
     def post_create(self):
         """Post create hooks"""
--- a/functionary/core/tests/models/test_function.py
+++ b/functionary/core/tests/models/test_function.py
@@ -1,0 +1,87 @@
+import pytest
+from django_celery_beat.models import CrontabSchedule, PeriodicTask
+
+from core.models import Function, Package, ScheduledTask, Task, Team, User
+
+
+@pytest.fixture
+def user():
+    return User.objects.create(username="user")
+
+
+@pytest.fixture
+def team():
+    return Team.objects.create(name="team")
+
+
+@pytest.fixture
+def environment(team):
+    return team.environments.get()
+
+
+@pytest.fixture
+def package(environment):
+    return Package.objects.create(name="testpackage", environment=environment)
+
+
+@pytest.fixture
+def function(package):
+    function_schema = {
+        "title": "test",
+        "type": "object",
+        "properties": {"prop1": {"type": "integer"}},
+    }
+    return Function.objects.create(
+        name="testfunction",
+        environment=package.environment,
+        package=package,
+        schema=function_schema,
+        active=True,
+    )
+
+
+@pytest.fixture
+def task(function, environment, user):
+    return Task.objects.create(
+        function=function,
+        environment=environment,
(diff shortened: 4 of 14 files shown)


functionary/ui/templates/core/function_detail.html
```diff
@@ -60,14 +70,58 @@ <h1 class="title is-1">
...
         {% if form is not None %}
-            <div class="field ml-4">
-                <label class="label">Create Task:</label>
-                <div class="ml-4">
-                    <form id="djangoForm"
-                          method="post"
-                          action="{% url 'ui:function-execute' %}">
-                        {% csrf_token %}
-                        <div class="has-background-white-ter">
+            {% if function.active %}
```
scott-taubman, line 73: Something definitely went wrong here because this template just fails to load.  You may want to just start with the template from main and re-apply your changes, rather than trying to sort out how to get it to rebase / merge properly.

functionary/builder/utils.py
```diff
@@ -303,11 +304,27 @@ def _create_functions_from_definition(definitions, package: Package):
...
+    db_functions = Function.objects.all()
+    for function_def in definitions:
+        name = function_def.get("name")
+        print(name)
+        try:
+            func = Function.objects.get(package=package, name=name)
+            db_functions.remove(func)
+        except Function.DoesNotExist:
+            pass
+    for inactive_functions in db_functions:
```
scott-taubman, line 324: Another minor quibble: you've named the variable `inactive_functions`, but it is 1) a singular function, not multiple, and 2) not actually inactive at this point in time.  I'd just call it `function`.

functionary/core/tests/models/test_function.py
```diff
@@ -0,0 +1,77 @@
...
+        periodic_task=periodic_task,
+    )
+
+
+@pytest.mark.django_db
+def test_deactivate(function):
+    """This marks the function to inactive and scheduled tasks to pause"""
+    function.deactivate()
+    assert not function.active
+    for scheduled_t in function.scheduled_tasks.all():
```
scott-taubman, line 76: You've somewhat similarly written this second part of the test in a way that it's easy to trick yourself into thinking it's passing, which you've done.

You're not actually loading the `scheduled_task` fixture.  So when you get here you're iterating over an empty set.  That means you're not actually testing the thing you're intending to test.  Another way you could possibly fool yourself here is if the scheduled tasks were already "PAUSED".  Then you've not verified that the deactivation process is what paused them.

What you can do is an upfront assertion to verify that there are scheduled tasks and they are currently active.  So at the front of your test, do:

```python
    assert function.active is True
    assert function.scheduled_tasks.filter(status=ScheduledTask.ACTIVE).exists()
```

Then you know that the assertions run after you call `function.deactivate()` are meaningful.

Also, note the use of the `ScheduledTask.ACTIVE` constant versus the string "ACTIVE".  Do the same thing for `PAUSED`.

functionary/ui/templates/core/task_detail.html
```diff
@@ -1,6 +1,18 @@
 {% extends "base.html" %}
 {% block content %}
     <div id="task_detail"
+        <div class="columns is-centered">
```
scott-taubman, line 4: I'm not sure what these changes are about.  I think you can just remove them.

functionary/core/models/function.py
```diff
@@ -43,6 +44,13 @@ class Function(models.Model):
     variables = models.JSONField(default=list, validators=[list_of_strings])
     return_type = models.CharField(max_length=64, null=True)
     schema = models.JSONField()
+    active = models.BooleanField(default=True)
+
+    def deactivate(self):
```
scott-taubman, line 49: Move this method to the bottom, below `__str__`.

scott-taubman (changes requested):

scott-taubman:
Late comment not related to a particular line of code:  You checked in the example/demo tarball, which we don't want to do.  Be sure to remove that file as well.

functionary/builder/utils.py
```diff
@@ -303,11 +304,27 @@ def _create_functions_from_definition(definitions, package: Package):
...
 
     return db_functions
 
 
+def _deactivate_functions(definitions, package: Package):
+    """Grabs all functions from DB, deactivates any that aren't in the definition"""
+    db_functions = Function.objects.all()
+    for function_def in definitions:
+        name = function_def.get("name")
+        print(name)
```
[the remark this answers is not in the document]
lights7412: Fixed.

functionary/core/migrations/0008_workflowparameter_and_more.py
```diff
@@ -65,6 +65,51 @@ class Migration(migrations.Migration):
             model_name="workflow",
             name="schema",
         ),
+        migrations.AddField(
```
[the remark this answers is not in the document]
lights7412: Started fresh from mains migration.

functionary/builder/utils.py
```diff
@@ -303,11 +304,27 @@ def _create_functions_from_definition(definitions, package: Package):
...
+        try:
+            func = Function.objects.get(package=package, name=name)
+            db_functions.remove(func)
```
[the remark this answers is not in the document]
lights7412: Used .exclude per recommendation

functionary/builder/utils.py
```diff
@@ -303,11 +304,27 @@ def _create_functions_from_definition(definitions, package: Package):
         function_obj.schema = _generate_function_schema(
             name, function_def.get("parameters")
         )
+        function_obj.active = True
         db_functions.append(function_obj)
 
     return db_functions
 
 
+def _deactivate_functions(definitions, package: Package):
```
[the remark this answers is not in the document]
lights7412: Tried to fix spacing

functionary/core/models/task.py
```diff
@@ -106,6 +106,11 @@ def clean(self):
         self._clean_environment()
         self._clean_parameters()
 
+    def validate_active(self):
+        active = self.function.active
+        if not active:
```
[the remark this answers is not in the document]
lights7412: Fixed.

functionary/core/tests/models/test_function.py
```diff
@@ -0,0 +1,77 @@
...
+        "type": "object",
+        "properties": {"prop1": {"type": "integer"}},
+    }
+    return Function.objects.create(
+        name="testfunction", package=package, schema=function_schema, active=True
+    )
+
+
+@pytest.fixture
+def task(function, environment, admin_user):
```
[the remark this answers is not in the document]
lights7412: Fixed to user

functionary/core/tests/models/test_function.py
```diff
@@ -0,0 +1,77 @@
...
+        parameters={"name": "input", "summary": "summary", "type": "text"},
+        periodic_task=periodic_task,
+    )
+
+
+@pytest.mark.django_db
+def test_deactivate(function):
+    """This marks the function to inactive and scheduled tasks to pause"""
+    function.deactivate()
+    assert not function.active
```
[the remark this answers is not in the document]
lights7412: fixed.

functionary/ui/forms/tasks.py
```diff
@@ -102,6 +103,10 @@ def _prepare_initial_value(param_type: str, initial: dict) -> Union[dict, None]:
     return None
 
 
+def get_available_functions(env: Environment) -> QuerySet[Function]:
```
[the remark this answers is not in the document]
lights7412: Got rid of this.

functionary/ui/templates/core/function_detail.html
lights7412: Started with main and reimplemented.

functionary/builder/utils.py
lights7412: Fixed to function

functionary/core/tests/models/test_function.py
lights7412: Did upfront assertion and used the constant, paused/active

functionary/core/models/function.py
lights7412: Fixed.

functionary/ui/templates/core/task_detail.html
lights7412: Fixed.

scott-taubman (changes requested):

scott-taubman (approved):

PR #177, opened 2023-01-24
#169 - Decreased scheduler wakeup sync delay

### LINKED ISSUE #169: Updating ScheduledTask parameters causes a delay in the task firing
From @xeonneon:

If you update a ScheduledTask, it will not run until the Schedulers next wakeup sync. For example, if I have a task that runs every minute and I update the params of that ScheduledTask, it won't run once the current minute ends. It will run during the next minute, after the Scheduler wakes up and resyncs. I think it's only an issue for tasks that run every minute.

xeonneon:
Closes #169 

## Introduced Changes
- Decreased Scheduler's WakeUp Sync Interval
  - This interval is how often the Scheduler queries the DB to figure out which `PeriodicTask`s have been updated. More specifically, which `PeriodicTask`s status' have been updated
  - The caveat with the solution is that if the problem will always exist, its just a matter of how often we want the Scheduler to re-sync with the DB. In this PR, I opted to let it re-sync every 5 seconds. So if I activate, pause, or archive the ScheduledTask when there is less than 5 seconds left before the task is scheduled to run, the task will follow whatever state it was prior to me submitting the updated status.

## Testing
Activate a previously paused or archived ScheduledTask and verify the ScheduledTask runs when expected

### DIFF
--- a/functionary/core/management/commands/run_scheduler.py
+++ b/functionary/core/management/commands/run_scheduler.py
@@ -13,6 +13,9 @@ logger = logging.getLogger("celery.beat")
 logger.setLevel(getattr(logging, LOG_LEVEL))
 
 
+CHECK_FOR_CHANGES_INTERVAL_SECONDS = 5
+
+
 class Command(BaseCommand):
     help = "Run the scheduler"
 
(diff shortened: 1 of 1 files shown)


scott-taubman (approved):
I worry that every 5 seconds might be a little spammy, but if it somehow ends up being a problem it's easy enough to dial it back.

PR #185, opened 2023-01-26
#182 - Added table template, added pagination to all list views

### LINKED ISSUE #182: Convert list views to tables
We've experimented with different ways to display data in the various list views.  We may come up with other ideas down the line, but for now we should just switch the remaining non-table pages and convert them to be tables.  This will simplify things when we tackle the css library switch and add filtering to the list views.

xeonneon:
Closes #182 

## Introduced Changes
- All list views now include pagination
- All list views now use a Table template, similar to the modals
- Added some small css consistency changes

## Testing
View all of the existing list views. Ensure pagination works as intended


functionary/ui/templates/partials/tasks/list.html
```diff
@@ -0,0 +1,21 @@
+{% extends "partials/table.html" %}
+{% block table_headers %}
+    <th>Created at</th>
```
scott-taubman, line 3: I think the order of these columns should go:

* Function
* Status
* Creator
* Created At

With the function name just linking over the task.  I don't know that there needs to be a link over to the function detail page from here.  We can probably find a convenient home for that link on the Task detail page eventually.

Same for the Builds table (Package > Status > Creator > Created At).

functionary/ui/templates/core/function_list.html
```diff
@@ -1,38 +1,5 @@
...
-                                </p>
-                                {% if function.summary %}<p class="subtitle is-6">{{ function.summary }}</p>{% endif %}
-                            </div>
-                        </div>
-                    </div>
-                </div>
-            </div>
-        {% endfor %}
-    </div>
+    {% include "partials/functions/list.html" %}
```
scott-taubman, line 3: What's the motivation behind the second layer here with the partial as opposed to just moving the contents of that partial here?  Is the thought that we might want to embed the full table in other places?  If so, where? If there isn't a compelling example use case I'd vote to remove the partials and just put the contents in each of these list templates directly.

scott-taubman (changes requested):

functionary/ui/templates/core/function_list.html
xeonneon: Since the tables now work the same as modals, you need to extend from them to fill out the table header and row blocks. With Django templates, you cannot extend from more than one file, so I either had to abandon the table partial, which would have been annoying to refactor in the future if we wanted to reformat the tables, or I could do what I did and just include a partial that extends from the table partial. Since we are planning on switching css libraries in the near future, I felt that the second option made more sense to make that refactor less painful

scott-taubman: I wonder if this is more of a custom template tag situation then...Carry on with how you have it organized for now and maybe we can play around with other ideas in the future.

scott-taubman (approved):

PR #188, opened 2023-01-27
#184 - __str__ now returns the display_name if set

### LINKED ISSUE #184: Fix display_name handling
At some point early on the whole point of the `display_name` on the functions and packages got lost, and we ended up with various incarnations of something like this throughout the code:

```
{% if package.display_name %}
    {{ package.display_name }}
{% else %}
    {{ package.name }}
{% endif %}
```

The idea behind display_name was supposed to be this:

* The `name` field is effectively a slug value; a unique identifier that is definable at creation time and friendlier to work with than the uuid.
* Since the `name` field is not allowed to change (if the name changed, the package or function would be treated as a new entity during the publish process), the `display_name` is the freely changeable alternative that allows people to change what gets shown as the name in the UI without touching the actual slug value.

In the vast majority of cases, we should be showing just the display_name if it exists.  If it doesn't exist, we show the name instead.  More importantly, that decision of what to show should be not be scattered throughout different templates or views.  Instead, the template code shown above should simply be something like:

```
{{ package.display_name }}
```

It's possible that the naming of the fields, `name` and `display_name`, inherently lend themselves to confusion since by the name alone it's not obvious what the distinction is.

One possible solution is to slightly rework the field names and then make `display_name` a property method helper.  Something like:

* `name` --> `slug`
* `display_name` --> `name`
* New property method `display_name` --> `return name or slug` 

We could keep the naming as is and come up with a new name for the property method, but `display_name` seems like the natural choice for that.

xeonneon:
Closes #184 

## Introduced Changes
- Updated Function and Package `__str__` method to return the `display_name` if it is set. Otherwise, return the `name`
  - This allows for the `name` and `display_name` to continue to exist as implemented. The templates just need to print out the `Function` or `Package` instead of the `name` or `display_name` attributes
- Updated the templates to display the `Function` or `Package` rather than the `name` or `display_name` attributes
- Removed some unnecessary `name` and `display_name` checks for teams and environments 

## Testing
Just need to visit the Function, Package, and Task views and verify the `Function` and `Package` names are still being displayed


scott-taubman (changes requested):
I don't love using `__str__` for this.  To me, it makes for some ambiguity when reading code that leverages it.  For example:

```python
 <span>{{ task.function }}</span>
```

Knowing that `task.function` is an object, what am I expecting to be rendered here?  Not even specifically, but generally?  Could be the id, the name, the `__repr__`...I basically have to go look at the `__str__` implementation to have any clue.  Or even to have confidence that thing that will get displayed matches the intent.

I think having a clearly defined property method makes both the intent and the expected outcome more clear to a reader.

Also, in at least one case here (the change to `utils/tasking.py`) you'd have to call `str()` to ensure the same behavior as before, and again `task.function.display_name` (or whatever the property method is called) feels more explicit than `str(task.function)`.

functionary/ui/templates/core/environment_detail.html
```diff
@@ -39,7 +39,7 @@ <h1 class="title is-1">
                 <ul id="packages">
                     {% for package in packages %}
                         <li class="block ml-4">
-                            <a href="{% url 'ui:package-detail' package.id %}">{{ package.name }}</a>
+                            <a href="{% url 'ui:package-detail' package.id %}">{{ package }}</a>
```
scott-taubman, line 42: missed a `package.render_name` here.

scott-taubman (changes requested):

functionary/ui/templates/core/environment_detail.html
xeonneon: Fixed

scott-taubman (approved):

PR #194, opened 2023-01-31
Reorganize views, templates, and urls for improved consistency

scott-taubman:
Reorganizing some things that have been irking me for a bit...

* Top level `templates` directory is now `package_templates` so that vscode django plugins don't try to treat the files as django templates.
* Took a pass at making the url names consistent, under the form `<model>-<verb>` (e.g. scheduledtask-create, workflow-update, etc.)
* Renamed the views files and directories to have singluar naming rather than plural so that they match the model file names.  For example, the model found under `models/package.py` has its views under `views/package.py` rather than `views/packages.py`.
* Nested urls kwargs that were of the form `<parent_model>_id` are now `<parent_model>_pk` to better align with the way the generic views expect the pathing to be.
* Moved the user_role management views under a subdirectory of the `team` and `environment` view folders since the old home implied actions against the environment / team rather than the user roles (e.g. environment/create.py --> environment/user_role/create.py)


kvig (approved):
Changes look good, everything still works.

Issue #183, opened 2023-01-25
Add environment to Function model


scott-taubman:
For most of the top level models that are permissioned by environment we've stamped the environment directly on the model for both ease of reference and query optimization (indexing can be done by environment).  Function is an outlier in this, so we should just go back and change that.  Things that need to happen here:

* Add environment to the function model
* Change the indexes to be two part indexes with environment as the first field (see Task as an example).
* Add a `clean()` check similar to what's on the Task model that ensures the environment of the function matches that of the package.
* Find any places where we're doing Function lookups via `package__environment` and switch those to filter on `environment` directly.

PR #195, opened 2023-02-01
#183 - Fix package build function deactivation

scott-taubman:
Fix to #183 

This fixes a logic bug during the package build that was resulting in all functions for packages other than the one being published getting deactivated.  Now the logic to find which functions to deactivate is properly scoped to the package being published instead.

## Test Instructions
* The unit test for function deactivation has been overhauled to properly capture what we're trying to test: Only functions for the package being published that are not present in the package.yaml get deactivated.  All others remain active.
* To manually test, publish a package, then publish another one.  The functions from the first published package should still be active even after the second one is published.

### DIFF
--- a/functionary/builder/tests/test_utils.py
+++ b/functionary/builder/tests/test_utils.py
@@ -97,8 +97,8 @@ def function4(package2):
 
 
 @pytest.fixture
-def definitions(function1, function2):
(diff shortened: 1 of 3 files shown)


kvig (approved):
Worked for me

PR #200, opened 2023-02-03
#144 - Added file parameter support for function detail

### LINKED ISSUE #144: File parameters
Functions should be able to have "file" type parameters that take a file upload as input.

Without overly complicating an initial implementation, we should keep in mind that we will likely need support for different file storage backends with deployments able to select between the available backends (the two most relevant probably being database and S3-compatible).  Even if our initial implementation only supports a single backend (S3-compatible), we should at least strive to not lock ourselves out of the potential of supporting others.

Other things that will need consideration:

* How the file location gets stamped into the tasking message and how we indicate that it is a file location (as opposed to just a text field that happens to have a URL in it).
* How the file is made available to the package container (probably [signed urls](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/PrivateContent.html)).
* Updating the various package templates to handle files if necessary (may not be if it's just a signed url).

xeonneon:
Closes #144 

## Introduced Changes
- Added MinIO container to `docker-compose.yml`
- Added `MinioInterface` class that utilizes the MinIO client API to create buckets, put files, delete files, check if files exist, etc.
- Added support for file parameters
- Task file parameters have their values substituted with presigned URLs before being sent to the runner
  - The presigned URLs are never saved to the database. They only need to exist in the task parameters when the task message is sent to the runner
- Added `multipart/form-data` encoding to all forms that spawn tasks
- Containers spawned by the runner now join the functionary network so the container can talk with MinIO

## Testing
- Publish the `file-param` package to Function
- Launch a task from the function detail view for the `Number of Characters` function
- Verify the function launches and displays the number of characters in the file


functionary/core/utils/minio.py
```diff
@@ -0,0 +1,95 @@
...
+        environment: The environment the request was made from
+        request: The HttpRequest that the form came from
+
+    Returns:
+        None
+    """
+    minio = MinioInterface(bucket_name=str(environment.id))
+    if request.FILES:
+        for _, file in request.FILES.items():
+            minio.put_file(file=file.file, length=file.size, filename=file.name)
```
scott-taubman, line 95: Using the file name alone here isn't going to be viable, since an environment can have any number of files uploaded that share a name.  Probably the thing to do is store it with a name that includes the task id and parameter name so that it's guaranteed to be unique.  So something like "<task_id>/<parameter_name>/<file_name>".

functionary/core/utils/parameters.py
```diff
@@ -0,0 +1,37 @@
+from collections import defaultdict, namedtuple
+
+Parameter = namedtuple("Parameter", "type format", defaults=(None, None))
+
+
+def parameter_mapping(schema: dict) -> dict[str, list[Parameter]]:
```
scott-taubman, line 6: It's worth pointing out that the parameter handling changes I'm making to get workflow and function parameters working the same is going to make this either change completely or go away entirely.  If other schema handling stuff comes up as part of this effort, don't spend any time trying to make it something wonderful, since it's all going to go away very soon.

functionary/ui/views/function.py
```diff
@@ -73,7 +74,9 @@ def execute(request: HttpRequest) -> HttpResponse:
 
     data: QueryDict = request.POST.copy()
     func = get_object_or_404(Function, id=data.get("function_id"))
-    form = TaskParameterForm(func, data)
+    handle_file_parameters(env, request)
```
scott-taubman, line 77: It seems like this ought to happen after the `form.is_valid()` call for sure.  Otherwise we're uploading files that may never get used because the other tasking input wasn't valid.

I might even say to put it after the `task.clean()` but before the `task.save()`.  That way the file upload is the last thing to happen before the tasking would actually get kicked off.

runner/runner/handlers.py
```diff
@@ -58,6 +58,7 @@ def _run_task(task):
             detach=True,
             command=run_command,
             environment=variables,
+            network_mode="functionary-network",
```
scott-taubman, line 61: Is this even a valid value?  Maybe you were intending to set `network`? I acknowledge that this works as is somehow, but the [documentation](https://docker-py.readthedocs.io/en/stable/containers.html) makes me think that's a happy accident.

Either way this is problematic and we'll need to figure out how to handle this.  For now, I guess I'd say to make this an environment variable that we pass to the runner and then set it to this value in our docker-compose.yml.

functionary/core/utils/minio.py
```diff
@@ -0,0 +1,95 @@
...
+from minio import Minio
+from urllib3.response import HTTPResponse
+
+from core.models import Environment
+
+
+class MinioInterface:
+    def __init__(self, bucket_name: str):
+        self.client = Minio(
+            endpoint=f"{getenv('MINIO_HOST')}:{getenv('MINIO_PORT')}",
```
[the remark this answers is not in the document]
scott-taubman: Late follow-up:  The settings should probably reference S3 rather than MINIO.  Also, I'm assuming that region would be required if using actual S3, so that should be an optional config setting here.

runner/runner/handlers.py
xeonneon: It should have just been `network="functionary-network"`. I'm honestly not sure why it lets me set `network_mode` to that value. Either way, the container that the runner spawns needs to be able to resolve the Minio container's hostname, so I figured I'd just add the container to the functionary network

scott-taubman: Right, I agree that this needs to be set to work in our dev environment, but you don't want to hard-code it in there.  It needs to be configurable so that it at least has a chance of working in some other setup.  In fact, we probably need both `network_mode` and `network` to be configurable and treated as optional.

examples/file-param/functions.py
```diff
@@ -0,0 +1,8 @@
+import requests
+
+
+def num_chars(file: str) -> int:
```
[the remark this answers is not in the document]
xeonneon: Fixed

functionary/ui/views/function.py
xeonneon: Fixed

functionary/core/utils/minio.py
xeonneon: Fixed. One thing that we might want to consider in the future is how Functionary admins will be able to clean up existing files. I imagine we could just come up with a UI page that lists all the files in the environment's bucket, and then delete files that fall within date ranges or some other filters

functionary/core/utils/minio.py
```diff
@@ -0,0 +1,95 @@
...
+            response.close()
+            response.release_conn()
+
+    def get_presigned_url(self, filename: str) -> str:
+        if self.bucket_exists():
+            return self.client.get_presigned_url(
+                method="GET",
+                bucket_name=self.bucket_name,
+                object_name=filename,
+                expires=timedelta(minutes=5),
```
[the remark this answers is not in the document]
xeonneon: Fixed

functionary/core/utils/minio.py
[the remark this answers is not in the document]
xeonneon: Fixed

functionary/core/utils/minio.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def get_filenames(self, filename: str) -> list[str]:
+        objects = self.client.list_objects(self.bucket_name)
+        object_names = []
+        for object in objects:
+            object_names.append(object.object_name)
+        return object_names
+
+    def does_file_exist(self, filename: str) -> bool:
+        objects = self.get_filenames(filename)
```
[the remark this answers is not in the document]
xeonneon: Fixed. I think I originally didn't use that method since it didn't return `None` if the file did not exist

runner/runner/handlers.py
xeonneon: Apparently `network_mode` and `network` are mutually exclusive options according to the documentation. Or at least that's what I believe it means when it says "incompatible"

scott-taubman: Yeah, I know you can't set both, but there are situations where you might want to set one or the other (as in our dev setup case) and we should provide a way to do that.

scott-taubman (commented):
Added some more, mostly minor comments.

One thing we'll want to do as a follow-on to this PR is add exception handling the appropriate places so that the user facing behavior in the event of a file upload failure is something appropriate.  Some unit tests verifying the behavior in failure cases would be good too.  That can happen as a separate PR though.

docker/docker-compose.yml
```diff
@@ -179,6 +183,20 @@ services:
       - 5000:5000
     volumes:
       - registry-data:/var/lib/registry
+  minio:
+    image: quay.io/minio/minio:RELEASE.2023-01-25T00-19-54Z
```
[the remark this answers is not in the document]
xeonneon: That's just the versioning scheme that MinIO uses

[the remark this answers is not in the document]
scott-taubman: I guess I'm suggesting that I'd rather just use the `latest` tag and if that breaks our stuff one day we'll fix it, rather than pinning to this super specific version (which is already not the newest release) and then likely staying on it for way too long.

functionary/functionary/settings/base/s3_.py
```diff
@@ -0,0 +1,14 @@
+from os import getenv
+
+S3_HOST = getenv("MINIO_HOST")
+S3_PORT = getenv("MINIO_PORT")
+S3_REGION = getenv("MINIO_REGION")
+S3_SECURE = False
+S3_ACCESS_KEY = getenv("MINIO_ROOT_USER")
+S3_SECRET_KEY = getenv("MINIO_ROOT_PASSWORD")
+S3_SIGNED_URL_TIMEOUT_MINUTES = 1
```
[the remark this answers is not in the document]
xeonneon: Fixed

functionary/functionary/settings/base/s3_.py
```diff
@@ -0,0 +1,14 @@
+from os import getenv
+
+S3_HOST = getenv("MINIO_HOST")
```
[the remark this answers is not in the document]
xeonneon: Fixed. I do have to leave the `MINIO_ROOT_USER` and `MINIO_ROOT_PASSWORD` as they are since those are special env vars just for MinIO

runner/runner/handlers.py
xeonneon: Fixed, although I don't check for true exclusive mutuality between network and network mode. I don't really see when we would want to modify the network mode. I figure that the task containers should always attach to the same network that the rest of the Functionary components are running on, as to not introduce a breaking configuration such that the task containers can no longer talk with Functionary's MinIO container

docker/docker-compose.yml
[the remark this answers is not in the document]
xeonneon: Fixed

PR #203, opened 2023-02-07
#186 - Add FunctionParameter model and reduce schema parsing

### LINKED ISSUE #186: Revisit Function Schema Storage and Validation
Early on we made the decision to store the Function definitions as an OpenAPI compatible schema on the Function model.  This was done largely with a couple of ideas in mind:

* The UI was going to be a separate application built on React or something similar, and serving the Function definitions up as a schema in a standard format made a good amount of sense.
* The builder might be or become a standalone service and again using OpenAPI as a standard way to define the functions made sense in terms of having the builder communicate to the core application what the functions looked like.

Neither of these original assumptions really came to fruition, most particularly the first one.

Somewhat separately from the above, we've failed to provide an abstraction layer for interacting with the schemas, which has lead to schema parsing code being sprinkled throughout the code base.  For example, the stuff going on [here](https://github.com/beer-garden/functionary/blob/main/functionary/core/utils/serialization.py#L8) and [here](https://github.com/beer-garden/functionary/blob/main/functionary/ui/forms/tasks.py#L62).

We absolutely need to create a class or other abstraction for interacting with the function schemas so that we don't have parsing code all over the place.  However...

We now also have the concept of Workflows, which in terms of how you task them, validate parameters, etc. are fundamentally the same as Functions.  But since Workflows are not created by the builder, there is no schema created for them inherently.  So the question becomes:

* Do we generate a schema for Workflows and then create an abstraction for interacting with the schemas?
* Or, do we switch the Functions over to store their definitions in a model similar to the Workflows and then come up with a solution for the parameter validation that does not rely on the OpenAPI compatible schemas?

My feeling leans toward that second option.  To be clear, we can still in the future generate schemas for Functions and Workflows if that is a thing we find value in doing.  It would just be that the schemas are not the actual mechanism for validation and the way we create them would be different (most likely no longer leveraging pydantic).

There are pros and cons to each approach and possibly some options that are not considered here.

scott-taubman:
Closes #186 

This switches the `Function` model to having its parameter definitions stored in a new model `FunctionParameter`, rather than through the pre-generated `schema`.  The reasons for doing this are:

* It makes `Function` and `Workflow` consistent in their parameter handling
* It makes interacting with the Function parameters (find parameters of type X) much easier, since wasn't any tooling for parsing and interacting with the pydantic generated schema.
* It makes us must less tied to the pydantic library and its schema format, since we have the parameter data in a format that we can easily feed into any library of our choosing.

The main changes that are happening here:

* A new `FunctionParameter` model now exists.  It and `WorkflowParameter` share a common base, so those two models and the base model now live in `core/models/parameter.py`.
* The builder has been updated to create (and delete) `FunctionParameter` instances rather than generating a pydantic schema.
* The `TaskParameterForm` now uses the info from the `FunctionParameter` instances to construct the form, rather than relying on the schema.
* There is now a `core/utils/parameter.py` (not to be confused with the now deleted `parameters.py`) that is the home of all of the logic that deals with pydantic.  Generating the schema and validating parameters against that schema all happen here.  This file is also now the home of the single source of parameter type definitions.  The various places that had defined their own parameter types now just import it from here.

Other related changes:
* The code that was in `core/utils/serialization.py` was no longer needed after these changes so that file is now gone.  A much simpler helper method that just finds the `json` parameters from the database is included in `core/utils/parameter.py` and used internally by t he `validated_parameters` function there.
* Many of the unit tests needed to be update to replace the function schema in various fixtures with `FunctionParameter` instances.

## Test Instructions
* Some new unit tests were written to confirm that the validation on `WorkflowRun` and the `FunctionParameter` removal during the build process work as expected.
* Make sure that package builds still work, functions are still taskable, and field validation in the UI still works.  The behavior should be the same as before in all cases.


functionary/builder/api/v1/serializers/package_definition.py
```diff
@@ -1,19 +1,10 @@
...
-    ("text", "String (Long)"),
-    ("float", "Float"),
-    ("file", "File"),
-    ("boolean", "Boolean"),
-    ("date", "Date"),
-    ("datetime", "Date Time"),
-    ("json", "JSON"),
-]
-RETURN_TYPES = PARAMETER_TYPES[:]
+RETURN_TYPE_CHOICES = PARAMETER_TYPE_CHOICES[:]
```
scott-taubman, line 7: I'm not entirely clear on why this is a separate thing from `PARAMETER_TYPE_CHOICES`, but that's how it was before and I decided to just keep it (though I renamed it for consistency).

functionary/builder/utils.py
```diff
@@ -191,16 +190,15 @@ def build_package(build_id: UUID):
...
-        # don't save it until the build has finished.
-        db_functions = _create_functions_from_definition(
-            package_definition.get("functions"), package
-        )
-        _deactivate_functions(package_definition.get("functions"), package)
-        _extract_package_contents(package_contents, workdir)
-        _load_dockerfile_template(dockerfile, workdir)
+    # Need to validate the potentially new function schema, but
+    # don't save it until the build has finished.
+    functions, function_parameters = _create_functions_from_definition(
```
scott-taubman, line 195: This got bumped outside of the try block since:

* The exceptions being handled didn't seem like they could be raised by the lines here.
* The linter was rightfully complaining later on that technically `functions` could be unset.

functionary/builder/utils.py
```diff
@@ -224,11 +222,18 @@ def build_package(build_id: UUID):
...
                 func.save()
+
+            for parameter in function_parameters:
+                parameter.save()
+
             package.image_name = image_name
             package.save()
 
+            _delete_removed_function_parameters(function_parameters, package)
+            _deactivate_removed_functions(function_definitions, package)
```
scott-taubman, line 235: I realized this `deactivate_removed_functions` call was in the wrong spot before, so I fixed that here.

functionary/core/models/function.py
```diff
@@ -80,7 +81,18 @@ def deactivate(self):
         for scheduled_task in self.scheduled_tasks.all():
             scheduled_task.pause()
 
+    @property
+    def parameters(self):
+        """Convenience alias for functionparameter_set"""
+        # Provides better static type checking than using related_name
```
scott-taubman, line 87: I started doing this here and will probably go through and do it for everything at some point.  Basically, when we use `related_name`, django sets that up dynamically at runtime, which means the static type checker doesn't know about it.  Explicitly defining it as a property method fixes that for the type checking and also makes it more obvious to someone looking at this model that the relationship is there.

functionary/ui/forms/tasks.py
```diff
@@ -33,18 +35,18 @@ class HTMLDateTimeInput(DateTimeInput):
...
 _field_mapping = {
-    "integer": (IntegerField, None),
-    "string": (CharField, None),
-    "text": (CharField, Textarea),
-    "number": (FloatField, None),
-    "file": (CharField, FileInput),
-    "boolean": (BooleanField, None),
-    "date": (DateField, HTMLDateInput),
-    "date-time": (
+    PARAMETER_TYPE.INTEGER: (IntegerField, None),
```
scott-taubman, line 38: I know one of the complaints has been that there are multiple mappings to go update when we add a parameter type.  I don't know that we can get it down to only 1 place, since in a case like this it's a mapping that's hyper specific to the TaskParameterForm.  I think having the constants defined at least makes it a little less error prone.

kvig:
I was testing and accidentally deleted part of a function and left the parameters section. The package.yaml ended up with two parameter sections:
```
      # (required) Parameters that the function takes
      parameters:
        - name: a
          type: integer
          default: 1
          required: true
        - name: b
          type: integer
          default: 1
          required: true

      # (required) Parameters that the function takes
      parameters: []
```
I don't know if this was a feature before the refactor, but if we can catch this misconfiguration that'd be good thing. I'd error on multiple parameter definitions.

scott-taubman:
> The package.yaml ended up with two parameter sections...if we can catch this misconfiguration that'd be good thing

We unfortunately cannot detect this particular error, since the library we're using (pyyaml) merges duplicate keys rather than throwing an error.  This goes against the yaml spec, but it seems like they've never reached consensus on how to fix it.  See discussions [here](https://github.com/yaml/pyyaml/issues/165) and [here](https://github.com/yaml/pyyaml/issues/41).

functionary/core/models/parameter.py
```diff
@@ -20,7 +13,43 @@
 )
 
 
-class WorkflowParameter(models.Model):
+class Parameter(models.Model):
+    """Base model for common components of FunctionParameter and WorkflowParameter"""
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
kvig, line 19: Need to update the urls.py, it's expecting the parameter id to be an int for workflow parameter edit and delete

kvig (changes requested):

functionary/core/models/function.py
scott-taubman: No, not really unfortunately.  It returns a RelatedManager which is dynamically generated based on the manager of the model class.

functionary/core/models/parameter.py
scott-taubman: Good catch.  Fixed.

functionary/core/utils/tasking.py
```diff
@@ -179,12 +179,13 @@ def _handle_file_parameters(task: Task) -> None:
...
     environment = task.environment
     parameters = task.parameters
-    schema: dict = task.function.schema
 
-    param_map = parameter_mapping(schema)
-    for param_name, param_meta in param_map.items():
-        param_formats = [_format for _, _format in param_meta]
-        if "uri" in param_formats:
+    for parameter in task.function.parameters.filter(
+        parameter_type=PARAMETER_TYPE.FILE
```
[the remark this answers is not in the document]
scott-taubman: Yeah, that was indeed doable.  Added that filter and then got rid of the subsequent check since it's now only operating on file parameters that are definitely in the task parameters.

functionary/core/models/function.py
```diff
@@ -43,7 +45,6 @@ class Function(models.Model):
     description = models.TextField(null=True)
     variables = models.JSONField(default=list, validators=[list_of_strings])
     return_type = models.CharField(max_length=64, null=True)
-    schema = models.JSONField()
```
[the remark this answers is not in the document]
scott-taubman: Fixed.

kvig (approved):

PR #207, opened 2023-02-09
197 centralized table rendering logic

### LINKED ISSUE #197: Centralize table rendering logic
We are currently writing custom table HTML in every individual list template (scheduledtask_list.html, function_list.html, etc).  This makes it very difficult to ensure a uniform look and feel for every table and laborious to make table styling changes.

We should instead centralize the table rendering, most likely using [django-tables2](https://github.com/jieter/django-tables2). That libary allows us to construct tables as classes similar to how native django Forms are constructed.  Through good use of a common base class and possibly a customized version of the Column class we should be able to make it much easier to have consistency across all of our tables.

django-tables2 also plays nice with django-filters, so we should look at implementing this before tackling #146.

west270:
Closes #197 

I changed the tables to use django-tables2 for all of the views that can be accessed from the navigation bar.

For testing the views the tables should still work as they did before the prior to the change, and also you should be able to click the column headers to sort the rows.


functionary/ui/tables/schedule_task.py
```diff
@@ -0,0 +1,44 @@
...
+    )
+    function = tables.Column(
+        linkify=lambda record: reverse(
+            "ui:function-detail", kwargs={"pk": record.function.id}
+        ),
+    )
+    last_run = tables.DateTimeColumn(
+        accessor="most_recent_task__created_at",
+        verbose_name="Last Run",
+        linkify=lambda record: reverse(
```
scott-taubman, line 22: This throws an exception if the task hasn't run yet.  Instead of using a lambda here, you should write a function (in this same file, above the class) that you point this to instead.  The function would return either this reverse value if `record.most_recent_task` is set, or `None`.

functionary/ui/tables/schedule_task.py
```diff
@@ -0,0 +1,44 @@
+import django_tables2 as tables
+from django.urls import reverse
+from django.utils.html import format_html
+
+from core.models.scheduled_task import ScheduledTask
+
+
+class ScheduledTaskTable(tables.Table):
```
scott-taubman, line 8: I'm noting this here only because the template is gone...

We lost the Create buttons on the Schedules and Workflows pages.  Those will have to have their own template and not use the default so that you can add the button back.

scott-taubman (changes requested):
In general I think this looks good.  I noted a couple of issues and one minor code reuse item.

We've already discussed some follow-ons and one will definitely have to be figuring out the best way to define the attrs in the Meta in a single place, whether that be a base Meta, a custom base Table class, or something in the table template.  It's okay to punt on that for now, just writing it down so that we remember to circle back to it.

functionary/ui/tables/schedule_task.py
```diff
@@ -19,21 +27,18 @@ class ScheduledTaskTable(tables.Table):
     last_run = tables.DateTimeColumn(
         accessor="most_recent_task__created_at",
         verbose_name="Last Run",
-        linkify=lambda record: reverse(
-            "ui:task-detail", kwargs={"pk": record.most_recent_task}
-        ),
-        format="N j, Y, g:i a",
+        linkify=lambda record: generateLastRunUrl(record),
```
scott-taubman, line 30: Minor thing that doesn't really need to change, but I'll point out that this could have just been `linkify=generateLasRunUrl`, since `linkify` takes a callable.

PR #211, opened 2023-02-09
#198 - Added util command to not crash while migrations get loaded

### LINKED ISSUE #198: Scheduler crashes if migrations aren't run rather than waiting
When running `./manage.py run_scheduler` the runner attempts to start before it confirms that the database is ready.  Unlike most of the other processes, which won't attempt to access the database until some action forces them to, the scheduler needs the database straight away.  These leads to the scheduler just crashing if, for instance, the migrations haven't been run yet.

What would be friendlier, especially in the developer container startup use case, is for the scheduler to wait for the database to be ready similar to what we do for the message broker connection.  One important distinction here is that "ready" is not just "can I connect to the database", but also "are the tables I'm going to need available."  This could be an independent check for a known needed table that we use as a proxy check or, a wrapper around the `scheduler.run()` call that catches the specific exception and then waits some amount of time before trying again.

xeonneon:
Closes #198 

## Introduced Changes
- Added a util command for the management commands that will attempt to run the given command
  - The command will be re-run on a delay if the necessary migrations have not been run

## Testing
- Clear your database volume
- Start functionary
  - The scheduler container should not be in a stopped state and claim that the necessary migrations have not been run yet
- Apply the migrations
- The scheduler container will start working as intended once it detects the migrations

### DIFF
--- a/functionary/core/management/commands/utils.py
+++ b/functionary/core/management/commands/utils.py
@@ -1,0 +1,29 @@
+from time import sleep
+from typing import Callable
+
+from django.db.utils import ProgrammingError
+
+SLEEP_DURATION = 2
+
+
+def run(command: Callable) -> None:
+    """Run the given the command
+
+    The given command will be executed to see if the necessary
+    database migrations have been run. If the migrations have not
+    been run, the command will be re-run after a delay.
+
+    Args:
+        command: The callable command to execute
+
+    Returns:
+        None
+    """
+    is_ready = False
+    while not is_ready:
+        try:
+            command()
+            is_ready = True
+        except ProgrammingError:
+            print("Migrations for the given command have not been run yet.")
+            sleep(SLEEP_DURATION)
--- a/functionary/core/management/commands/utils.py
+++ b/functionary/core/management/commands/utils.py
@@ -1,9 +1,15 @@
+import logging
 from time import sleep
 from typing import Callable
 
-from django.db.utils import ProgrammingError
+from django.conf import settings
+from django.db.utils import DatabaseError
 
-SLEEP_DURATION = 2
+LOG_LEVEL = settings.LOG_LEVEL
+logger = logging.getLogger(__name__)
+logger.setLevel(getattr(logging, LOG_LEVEL))
+
+SLEEP_DURATION = 5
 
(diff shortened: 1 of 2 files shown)


functionary/core/management/commands/utils.py
```diff
@@ -0,0 +1,29 @@
...
+    Returns:
+        None
+    """
+    is_ready = False
+    while not is_ready:
+        try:
+            command()
+            is_ready = True
+        except ProgrammingError:
+            print("Migrations for the given command have not been run yet.")
```
scott-taubman, line 28: Two things with this:

1) Let's use the logger rather than a print statement.  Minor thing, but as is this ends up showing as a WARN rather than an ERROR, which you should be able to fix by calling `logger.error` instead.

2) Let's make the wording more general so that it applies to general database readiness issues:
  >Database is not ready. Ensure database is available and migrations have been run.  Retrying in 
  {SLEEP_DURATION}s.

You could break out the errors so that the message is specific to the actual problem, but the stack trace ends up being right above the error message anyway.

functionary/core/management/commands/utils.py
```diff
@@ -0,0 +1,29 @@
+from time import sleep
+from typing import Callable
+
+from django.db.utils import ProgrammingError
+
+SLEEP_DURATION = 2
```
scott-taubman, line 6: Let's go with 5.  I don't know why, but 2s feels like it's below my "too spammy" threshold for this.

xeonneon: Fixed

functionary/core/management/commands/utils.py
xeonneon: Fixed

functionary/core/management/commands/utils.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Returns:
+        None
+    """
+    is_ready = False
+    while not is_ready:
+        try:
+            command()
+            is_ready = True
+        except ProgrammingError:
```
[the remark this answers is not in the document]
xeonneon: Fixed. Opted for `DatabaseError`

PR #212, opened 2023-02-09
#204 - Added S3 error handling to function detail view

### LINKED ISSUE #204: Improve File Upload Error Handling
When tasking a function that takes a file parameter a 500 error gets returned and the there is no user friendly handling of that response.  The proper response code is probably 503 (Service Unavailable), but more importantly we should treat the error like a form error and display an error similar to how the non-field errors get displayed informing the user that there was an issue receiving their file and that they should try again later.

The easiest way to replicate this is to try uploading a file without the s3 container running, or mis-configuring the s3 credentials.  The specific error does not really matter though, we need to generically handle any situation where the file fails to make it into the bucket.

xeonneon:
Closes #204 

## Introduced Changes
- Throw appropriate exceptions when initializing the MinIO Client
  - `MaxRetryError` when the client is unable to connect to the MinIO instance
  - `S3Error` when client is unable to authenticate the with the MinIO instance
- Since you cannot use the MinIO client without first initializing the client, then this should prevent downstream client exceptions when a connection to MinIO cannot be established

## Testing
- Change the S3 Access Key or Secret key to something other than the Access Key and Secret Key that establishes the MinIO Root User
- Start Functionary
- Use the `Number of Files` function and try to submit a file
  - A non-field error should appear claiming that functionary was unable to authenticate with the S3 provider
- Stop the MinIO container and re-submit the file
  - The page should be in a loading state for about 5 seconds and a non-field error should appear claiming that functionary was unable to connect to the S3 provider

### DIFF
--- a/functionary/core/utils/minio.py
+++ b/functionary/core/utils/minio.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import timedelta
 from io import BytesIO
 
@@ -10,8 +11,36 @@ from urllib3.response import HTTPResponse
 
 from core.models import Task
 
+LOG_LEVEL = settings.LOG_LEVEL
+logger = logging.getLogger(__name__)
+logger.setLevel(getattr(logging, LOG_LEVEL))
+
+
+class S3ConnectionError(Exception):
+    _err_msg = (
+        "Unable to upload file. Please try again. "
+        "If the problem persists, contact your system administrator."
+    )
+
+    def __init__(self, msg: str = _err_msg, *args, **kwargs):
+        super().__init__(msg, *args, **kwargs)
+
 
 class MinioInterface:
+    """Wrapper class around MinIO API operations
+
+    This interface provides simplified functions for interacting with
+    MinIO, or any other S3 API compatible provider.
+
+    Attributes:
+        client: The MinIO client that communicates with the S3 provider
+        bucket_name: The name of the bucket the client will interact with
+
+    Raises:
+        S3ConnectionError: Raised when client is unable to connect or communicate
+            with the configured S3 provider.
+    """
+
     def __init__(self, bucket_name: str):
         try:
             self.client = Minio(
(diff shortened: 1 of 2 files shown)


functionary/core/utils/minio.py
```diff
@@ -5,23 +5,40 @@
...
+                secret_key=settings.S3_SECRET_KEY,
+                secure=settings.S3_SECURE,
+                region=settings.S3_REGION,
+            )
 
-        self.bucket_name = bucket_name
-        self._create_bucket()
+            self.bucket_name = bucket_name
+            self._create_bucket()
+        except MaxRetryError as err:
```
scott-taubman, line 27: I would catch these here and re-raise a custom exception.   Basically, if I'm instantiating this somewhere (like in the view code below) catching a urlib3 exception or a Minio exception based on the particular error seems like a weird implementation detail.  Instead, let's just define a custom exception at the top of this file and catch and raise.  Also, we should log the error:

```python
except MaxRetryError:
    msg = f"Connection to S3 provider at {settings.S3_HOST}:{settings.S3_PORT} could not be established"
    logger.error(msg)
    raise S3ConnectionError(msg)
except S3Error:
    # similar deal raising same S3ConnectionError
```

This also makes me realize that there's no docstring to this class.  We should add one with the usual class docstring info, plus info about the exception that may be raised.

functionary/ui/views/function.py
```diff
@@ -102,6 +104,18 @@ def execute(request: HttpRequest) -> HttpResponse:
...
+                ),
+            )
+        except S3Error as err:
+            form.add_error(
+                None,
+                ValidationError(f"{err.message}.", code="invalid"),
+            )
 
     args = {"form": form, "function": func}
     return render(request, "core/function_detail.html", args)
```
scott-taubman, line 121: We should update this to include a status code and then set the status code appropriately in the exception handling.  It's a little weird to run into the S3 related errors, for example, but have the HTTP response still be 200.  For ValidationError the response should be 400.  For the S3 errors I'd probably say 503.

scott-taubman (changes requested):

functionary/ui/views/function.py
xeonneon: Fixed, but just for the Function detail view. I'll add the status codes to the other views that spawn Tasks as I add support for file parameters for those views

functionary/core/utils/minio.py
xeonneon: Added the docstring and custom exception

functionary/ui/views/function.py
```diff
@@ -102,6 +104,18 @@ def execute(request: HttpRequest) -> HttpResponse:
                     code="invalid",
                 ),
             )
+        except MaxRetryError as err:
```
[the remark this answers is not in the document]
xeonneon: If I interpreted this right, I made the error message that gets rendered to the user be the generic nonsense error, while the error message that gets logged be the more specific error message.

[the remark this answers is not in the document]
scott-taubman: Yeah, it was close.  The generic user facing error really only belongs in the view code.  The exception raised inside MinioInterface should be the developer / system administrator relevant one.  I just went ahead and tweaked it.  Otherwise it all looks good.

scott-taubman (approved):

PR #214, opened 2023-02-10
Allow scheduling functions without parameters.

### LINKED ISSUE #196: Functions without parameters cannot be scheduled
It is currently not possible to create a scheduled task against a function that takes no parameters.  Clicking "Create" just takes you back to the scheduled task creation screen and no actual scheduled task is created.

kvig:
Closes #196 

This PR allows Functionary to schedule functions that don't have parameters.  It also updates the ScheduledTask model to create PeriodicTasks with the ScheduledTask's id instead of name and adds an edit button the the ScheduledTask details page.

Testing:
Try create a scheduled task that takes JSON input. Put invalid JSON in the field so validation fails.  Check that there is only one error shown at the field instead of at the top of the form.
Create another similar task with invalid JSON, but also add an invalid character for one of the cron fields. Ensure both errors are shown on the form and nothing at the top. When you correct the JSON, verify that the cron error is still shown.

### DIFF
--- a/functionary/core/models/scheduled_task.py
+++ b/functionary/core/models/scheduled_task.py
@@ -49,7 +49,7 @@ class ScheduledTask(ModelSaveHookMixin, models.Model):
         to="Function", on_delete=models.CASCADE, related_name="scheduled_tasks"
     )
     environment = models.ForeignKey(to="Environment", on_delete=models.CASCADE)
-    parameters = models.JSONField(encoder=DjangoJSONEncoder)
+    parameters = models.JSONField(encoder=DjangoJSONEncoder, blank=True)
     status = models.CharField(max_length=16, choices=STATUS_CHOICES, default=PENDING)
     creator = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.PROTECT)
     created_at = models.DateTimeField(auto_now_add=True)
@@ -150,7 +150,8 @@ class ScheduledTask(ModelSaveHookMixin, models.Model):
         """
         if self.periodic_task is None:
             self.periodic_task = PeriodicTask.objects.create(
-                name=self.name,
+                name=self.id,
+                description=self.description or self.name,
                 crontab=crontab_schedule,
                 task="core.utils.tasking.run_scheduled_task",
                 args=f'["{self.id}"]',
--- a/functionary/core/models/scheduled_task.py
+++ b/functionary/core/models/scheduled_task.py
@@ -49,7 +49,7 @@ class ScheduledTask(ModelSaveHookMixin, models.Model):
         to="Function", on_delete=models.CASCADE, related_name="scheduled_tasks"
     )
     environment = models.ForeignKey(to="Environment", on_delete=models.CASCADE)
-    parameters = models.JSONField(encoder=DjangoJSONEncoder, blank=True)
+    parameters = models.JSONField(encoder=DjangoJSONEncoder, blank=True, default=dict)
     status = models.CharField(max_length=16, choices=STATUS_CHOICES, default=PENDING)
     creator = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.PROTECT)
     created_at = models.DateTimeField(auto_now_add=True)
@@ -93,6 +93,8 @@ class ScheduledTask(ModelSaveHookMixin, models.Model):
 
     def _clean_parameters(self):
         """Validate that the parameters conform to the function's schema"""
+        if self.parameters is None:
+            self.parameters = {}
         validate_parameters(self.parameters, self.function)
 
     def clean(self):
--- a/functionary/core/tests/utils/test_parameter.py
+++ b/functionary/core/tests/utils/test_parameter.py
@@ -1,7 +1,7 @@
 import pytest
 from django.core.exceptions import ValidationError
 
-from core.models import Function, FunctionParameter, Package, Task, Team
+from core.models import Function, FunctionParameter, Package, Team
 from core.utils.parameter import PARAMETER_TYPE, validate_parameters
 
 
@@ -30,7 +30,7 @@ def function(package):
 
 
 @pytest.fixture
-def param1(function):
+def json_param(function):
     return FunctionParameter.objects.create(
         name="json_param",
         function=function,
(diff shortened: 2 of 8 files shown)


functionary/ui/forms/scheduled_task.py
```diff
@@ -11,7 +13,26 @@
 from core.models import Environment, Function, ScheduledTask
 
 
+class NotNoneJSONField(JSONField):
```
scott-taubman, line 16: I see how you arrived at this and I'll admit that the behavior of the form validation and handling of the empty value for parameters threw me for a loop when trying to understand what's going on here.  But this definitely feels like overkill and I don't think we should need to resort to this.

I tried playing with the `ScheduledTask` and the validation process to see if I could fix this more simply, and I think maybe doing the following gets us the behavior that we want:

* Updated the `ScheduledTask` parameters field to have `blank=True, default=dict`.  The blank part allows the form validation to pass when there are no parameters and the default means that when the instance is saved the parameters will be `{}` instead of `None`.
* Handle parameters=None in the `validate_parameters` util function:
   ```python
    if parameters is None:
        parameters = {}
    ```

That feels like a more natural solution than what's going on here.  If there are other scenarios that your solution is covering that what I've described above would not let me know.  But I'm hopeful that worst case is we could iterate on what's above and get there.

functionary/ui/forms/scheduled_task.py
```diff
@@ -52,11 +74,20 @@ class Meta:
...
+        environment: Environment = None,
+        existing_errors=None,
+        *args,
+        **kwargs,
+    ):
         super().__init__(*args, **kwargs)
         self._setup_field_choices(kwargs.get("instance") is not None)
         self._update_function_queryset(environment)
         self._setup_field_classes()
+        self.existing_errors = (
```
scott-taubman, line 88: I'm not in love with this `existing_errors` things.  It seems like it would be more straightforward to just handle the case where  `task_parameter_form.is_valid()` returns `False` in the `ScheduledTaskCreateView`.  Right now it does:

```python
if task_parameter_form.is_valid():
   pass
```

and then proceeds to still construct the `ScheduledTaskForm` and all that.  I think that's why you end up with both the form field error and non-form-field errors.  You could instead introduce handling and do something like:

```python
if not task_parameter_form.is_valid():
    context = {
        "form": self.get_form(),
        "task_parameter_form": task_parameter_form,
    }
    return render(
        request, "forms/scheduled_task/scheduled_task_edit.html", context
    )
```

That creates some minor duplication with what gets returned at the end of that method, so there's probably some room for refactor there to reduce that, but I feel like doing something like that to handle the `TaskParameterForm` errors before generating moving on to do the `ScheduledTaskForm` handling would be cleaner.

scott-taubman (changes requested):

functionary/core/utils/parameter.py
```diff
@@ -96,6 +96,7 @@ def validate_parameters(parameters: dict, instance: Union["Function", "Workflow"
     """
     try:
         pydantic_model = _get_pydantic_model(instance)
+        # KeyError will get thrown if one or more parameters are invalid
```
scott-taubman, line 99: I'm pretty sure this is actually a bug that needs to be fixed rather than just catching the `KeyError`.  `_serialize_json_parameters` assumes that all of the JSON parameters that a function takes are present in the parameters input, which is obviously not always going to be the case.  It should have `name__in=parameters.keys()` added to the filter so that it's only attempting to serialize the parameters that are actually present.  Then you would end up with a `ValidationError` as desired, rather than this `KeyError`.

Side note: we should absolutely be writing unit tests for situations like this.  We write a test to reproduce the bug, fix the bug, see the test pass.  This is technically not related to the thing that this PR is about, so if you want to take this change out and open a separate ticket for this bug I'm okay with that.  In fact, maybe it'd be better to do that because I actually want to talk over some of what's going on here with the model validation as I don't particularly love certain aspects of it.

functionary/core/models/scheduled_task.py
```diff
@@ -150,7 +152,8 @@ def set_schedule(self, crontab_schedule: CrontabSchedule) -> None:
         """
         if self.periodic_task is None:
             self.periodic_task = PeriodicTask.objects.create(
-                name=self.name,
+                name=self.id,
+                description=self.description or self.name,
```
scott-taubman, line 156: What's the motivation for adding the description here?  I don't see it becoming something that we would surface anywhere, as the `PeriodicTask` is an internal celery-beat thing that we're leveraging and not a user facing object.  It doesn't hurt anything to set it, I'm just curious if it there was something that lead you to do that.

functionary/core/utils/parameter.py
```diff
@@ -96,6 +100,8 @@ def validate_parameters(parameters: dict, instance: Union["Function", "Workflow"
     """
     try:
         pydantic_model = _get_pydantic_model(instance)
-        _ = pydantic_model(**_serialize_json_parameters(parameters, instance))
+        validate(
```
scott-taubman, line 103: Silly as it seems, your commented out test could pass if you use both the pydantic and jsonschema validation here, which you can do by calling `.schema()` on an instance of the model rather than the model class and then catching both types of `ValidationError`:

```python
    try:
        pydantic_model = _get_pydantic_model(instance)
        serialized_parameters = _serialize_json_parameters(parameters, instance)

        jsonschema.validate(
            serialized_parameters, pydantic_model(**serialized_parameters).schema()
        )
    except (ValidationError, jsonschema.ValidationError, json.JSONDecodeError) as exc:
        raise DjangoValidationError(exc)
```

functionary/core/tests/utils/test_parameter.py
```diff
@@ -0,0 +1,104 @@
...
+    with pytest.raises(ValidationError, match=r".*json_param.*is a required.*"):
+        validate_parameters({}, task.function)
+
+    with pytest.raises(ValidationError, match=r".*json_param.*is a required.*"):
+        validate_parameters({"true": False}, task.function)
+
+
+@pytest.mark.django_db
+@pytest.mark.usefixtures("param1", "param2")
+def test_incorrect_parameters2(task):
```
scott-taubman, line 97: A slight tweak to the naming and usage of the fixtures could make these tests easier to read and understand in isolation.  For example:

```python
@pytest.mark.django_db
def test_incorrect_parameter_type(task, json_param, date_param):
    """Check that parameters with an incorrect type don't work."""
    with pytest.raises(ValidationError, match=r".*date_param.*"):
        validate_parameters({json_param.name: 1, date_param.name: False}, task.function)

    with pytest.raises(ValidationError, match=r".*date_param.*"):
        validate_parameters(
            {json_param.name: 1, date_param.name: "False"}, task.function
        )
```

I feel like it's more obvious reading the above that `date_param` is being provided as the incorrect type, as where with `dont_hide` I have to go look at the fixture.

One other minor note is that it doesn't look like you're using the task fixture itself anywhere.  You could just use the function fixture directly and do away with the task.

PR #215, opened 2023-02-13
#213 - Added file param support for the API

### LINKED ISSUE #213: Add file parameter support to the API

xeonneon:
Closes #213 

## Introduced Changes
- The `/api/v1/tasks` endpoint now handles file input in the same way as the UI
  - The only caveat is that you cannot submit Tasks that rely on file parameters through the Swagger UI

## Testing
- Publish the demo package to Functionary
- Fetch an authentication token from the API
  - `curl -X POST http://localhost:8000/api/v1/api-token-auth -H 'accept: application/json' -F "username=" -F "password="`
- Submit a task to the `num_chars_api` test function
  - `curl -X POST http://localhost:8000/api/v1/tasks/ -H 'accept: application/json' -H 'X-Environment-Id: <env uuid>' -H 'Content-Type: multipart/form-data' -F "function=<function uuid>" -F "file.function=@/path/to/file" -F 'parameters={"other_param": "hello world"}' -H "Authorization: Token <your auth token>"`
- Verify task runs as expected

### DIFF
--- a/examples/demo/package.yaml
+++ b/examples/demo/package.yaml
@@ -65,18 +65,6 @@ package:
           description: Input file that will be measured
           type: file
           required: true
-    - name: num_chars_api
-      summary: Returns the number of characters in given file
-      display_name: Number of Characters API test
-      description: Returns the number of characters in given file
-      return_type: integer
-      parameters:
-        - name: function
-          display_name: Function
-          summary: File to measure
-          description: Input file that will be measured
-          type: file
-          required: true
         - name: other_param
           display_name: Other Param
           summary: Other param argument
--- a/examples/demo/package.yaml
+++ b/examples/demo/package.yaml
@@ -65,6 +65,24 @@ package:
           description: Input file that will be measured
           type: file
           required: true
+    - name: num_chars_api
+      summary: Returns the number of characters in given file
+      display_name: Number of Characters API test
+      description: Returns the number of characters in given file
+      return_type: integer
+      parameters:
+        - name: function
+          display_name: Function
+          summary: File to measure
+          description: Input file that will be measured
+          type: file
+          required: true
+        - name: other_param
+          display_name: Other Param
+          summary: Other param argument
+          description: Prints parameter
+          type: text
+          required: true
     - name: parameter_types
       summary: Test for parameter type form rendering and passthrough values
       display_name: Parameter Types
--- a/functionary/core/api/v1/utils.py
+++ b/functionary/core/api/v1/utils.py
@@ -64,6 +64,9 @@ def _cast_json_parameters(values: OrderedDict) -> None:
 
     for json_param in json_params:
         try:
+            if not values["parameters"].get(json_param.name):
+                continue
+
             if type(values["parameters"][json_param.name]) == dict:
                 continue
 
--- a/functionary/core/api/v1/utils.py
+++ b/functionary/core/api/v1/utils.py
@@ -4,6 +4,11 @@ from collections import OrderedDict
 from typing import Union
 
 from django.core.files.uploadedfile import InMemoryUploadedFile
+from rest_framework import serializers
+from rest_framework.exceptions import ValidationError
+
+from core.models import Function, FunctionParameter
+from core.utils.parameter import PARAMETER_TYPE
 
 PREFIX = "param"
 SEPARATOR = r"\."
@@ -17,7 +22,7 @@ def parse_parameters(values: OrderedDict) -> None:
 
     # Avoid dictionary changed size error by wrapping items in list
     for param, _ in list(values.items()):
-        if not (valid_param := r.match(param)):
+        if not (valid_param := get_parameter_name(param)):
             continue
 
         value = values.pop(param)
@@ -27,8 +32,78 @@ def parse_parameters(values: OrderedDict) -> None:
         values["parameters"][valid_param.group(2)] = value
 
 
+def cast_parameters(values: OrderedDict) -> None:
+    """Parent method to cast parameter necessary types"""
+    _cast_json_parameters(values)
+
+
 def get_parameter_name(parameter: str) -> Union[re.Match, None]:
+    """Return match if parameter name is of valid format"""
     return r.match(parameter)
+
+
+def _cast_json_parameters(values: OrderedDict) -> None:
+    """Mutates all JSON parameter fields into python objects
+
+    Args:
+        values: An OrderedDict containing the values passed to the serializer
+
+    Returns:
+        None
+
+    Raises:
+        ValidationError: When a JSON parameter is not valid JSON
+    """
+    function = _get_function(values)
+
+    json_params: list[FunctionParameter] = [
+        param
+        for param in function.parameters.all()
+        if param.parameter_type == PARAMETER_TYPE.JSON
+    ]
+
+    for json_param in json_params:
+        try:
+            if type(values["parameters"][json_param.name]) == dict:
+                continue
+
+            values["parameters"][json_param.name] = json.loads(
+                values["parameters"][json_param.name]
+            )
+        except json.decoder.JSONDecodeError as err:
+            exception_map = {json_param.name: err.msg}
+            exc = ValidationError(exception_map)
+            raise serializers.ValidationError(serializers.as_serializer_error(exc))
+
+
+def _get_function(values: OrderedDict) -> Function:
+    """Get the function from the internal values
+
+    Fetch the Function object from the values dictionary. If the `function` key
+    is not present, then we know the function name and package name serializer
+    was used. Since that serializer does not substitute the Function before
+    it's create method, we must fetch the function ourselves.
+
+    Args:
+        values: An ordered dict containing the values passed to the serializer
+
+    Returns:
+        function: The function object
+
+    Raises:
+        ValidationError: If the function was not found
+    """
+    try:
+        if not (function := values.get("function")):
+            function: Function = Function.objects.get(
+                name=values.get("function_name"),
+                package__name=values.get("package_name"),
+            )
+        return function
+    except Function.DoesNotExist as err:
+        exception_map = {"function_name": err}
+        exc = ValidationError(exception_map)
+        raise serializers.ValidationError(serializers.as_serializer_error(exc))
 
 
 def _update_parameter_field_type(values: OrderedDict) -> None:
--- a/functionary/core/api/v1/utils.py
+++ b/functionary/core/api/v1/utils.py
@@ -33,7 +33,7 @@ def parse_parameters(values: OrderedDict) -> None:
 
 
 def cast_parameters(values: OrderedDict) -> None:
-    """Parent method to cast parameter necessary types"""
+    """Parent method to cast parameters to necessary types"""
     _cast_json_parameters(values)
 
 
--- a/functionary/core/api/v1/utils.py
+++ b/functionary/core/api/v1/utils.py
@@ -13,7 +13,7 @@ from core.utils.parameter import PARAMETER_TYPE
 PREFIX = "param"
 SEPARATOR = r"\."
 
-r = re.compile(rf"^({PREFIX}){SEPARATOR}(\w+)$")
+param_name_format = re.compile(rf"^({PREFIX}){SEPARATOR}(\w+)$")
 
 
 def parse_parameters(values: OrderedDict) -> None:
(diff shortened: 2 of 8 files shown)


functionary/core/api/v1/views/task.py
```diff
@@ -67,8 +80,22 @@ def get_serializer_class(self):
         },
         parameters=HEADER_PARAMETERS,
     )
-    def create(self, request, *args, **kwargs):
-        request_serializer = self.get_serializer(data=request.data)
+    def create(self, request: Request, *args, **kwargs):
+        if not request.data.get("parameters"):
```
scott-taubman, line 84: This changes parameters from a required to an optional field, which may or may not have been the intent, but either way that change is not reflected in the endpoint documentation.  I'd personally just keep it as required for the sake of implementation simplicity and explicitness in execution.

functionary/core/api/v1/views/task.py
```diff
@@ -67,8 +80,22 @@ def get_serializer_class(self):
...
+            request.data["parameters"] = {}
+        else:
+            request.data["parameters"] = json.loads(request.data["parameters"])
```
scott-taubman, line 87: This will throw an exception when the content type is `application/json` because the data is already a dict.  As mentioned elsewhere, I'm okay with making this endpoint support `multipart/form-data` only, but if we're going to do that we need to update it to actually do that so it will return an appropriate response if someone tries to submit as `application/json`.  Also, the documentation needs to updated to reflect the supported content-type(s).

And with all that said, it seems like you it should be possible to have the serializers handle this for you rather than calling `json.loads` manually.

functionary/core/api/v1/views/task.py
```diff
@@ -67,8 +80,22 @@ def get_serializer_class(self):
...
+
+        # Wrap items in list to avoid dictionary changed size error
+        for param_name, _ in list(request.FILES.items()):
+            _handle_file_parameters(request, param_name)
+
+        request.data["parameters"] = json.dumps(request.data["parameters"])
```
scott-taubman, line 93: This further confuses me because you `json.loads` this data a few lines earlier, only to `json.dumps` it again here...

functionary/core/api/v1/views/task.py
```diff
@@ -77,6 +104,10 @@ def create(self, request, *args, **kwargs):
         )
 
         response_serializer = TaskCreateResponseSerializer(request_serializer.instance)
+        if request.FILES:
+            task = Task.objects.get(id=response_serializer.instance.id)
+            handle_file_parameters(task, request)
```
scott-taubman, line 109: It seems like maybe the fact that this takes a request and that the serializer normally takes in request.data was influencing some of your implementation decisions.  Meaning, that it looks like you're possibly trying to manipulate the different data structures in a way that keeps them intact for when you have to pass them in to things later on.  There's nothing that says the serializer needs to take `request.data` directly as opposed to some other dict and it's possible that slightly refactoring `handle_file_parameters` so that it similarly operates on an arbitrary dict as opposed to the request.FILES may prove helpful here.  Maybe not, but don't feel like you have to go out of your way to ensure that `request` and `request.data` look the way you need them to for these calls, when you could just be passing in some other object that you've separately constructed to be the way you need it.

functionary/core/api/v1/views/task.py
```diff
@@ -52,7 +58,14 @@ def get_serializer_class(self):
         description=(
             "Execute a function. The function to be executed can be defined either "
             "by supplying function as a string uuid, or function_name and "
-            "package_name."
+            "package_name. "
+            "In order to submit Tasks that take in file parameters, you must submit "
+            "requests to the API endpoint through the command line, and prefix the "
+            "file parameters with `file.`. The file parameters should not be placed "
+            "in the parameters argument. For example, "
+            "`-F 'function=1e8a7097-33b6-4ec4-87fc-52b7079caa76' "
```
scott-taubman, line 66: This documentation is not intended as instructions on how to use a specific tool to call the API, but rather documentation of how the endpoint works and what input it takes.  Basically you just need to describe the nature the of the `file.<parameter_name>` convention here and not how someone might use curl to call it.

examples/demo/package.yaml
```diff
@@ -63,5 +63,23 @@ package:
           display_name: File
           summary: File to measure
           description: Input file that will be measured
-          type:  file
+          type: file
+          required: true
+    - name: num_chars_api
```
scott-taubman, line 68: I don't think you need a separate function here.  You could just update the existing "num_chars" to take the additional parameter.

scott-taubman: Bonus points for making the additional parameter do something meaningful as it relates to the file so that it's a slightly more meaningful demo.

functionary/core/api/v1/views/task.py
xeonneon: Fixed. The original intention was that if a user passed a file parameter to a function that only accepts a file parameter, then the parameter field would need to be created if it did not already exist

functionary/core/api/v1/views/task.py
xeonneon: Fixed. The `json.loads` and `json.dumps` now only gets run if `request.FILES` is not `None`

functionary/core/api/v1/views/task.py
xeonneon: The `json.loads` was needed to turn the json string into a dict so I could manipulate the `parameters` field. Then the `parameters` field is supposed to be in a valid json string, so I `json.dumps` it afterwards

examples/demo/package.yaml
xeonneon: Agreed. Fixed, although I dropped the duplicate `function` parameter name, since I figured that would be confusing as a demo for a file parameter

xeonneon:
# New Changes:
- All parameters, including file parameters, can be specified with `param.<param_name>`
- You can now mix-and-match the way you write parameters
  - You can write out parameters as both `-F param.<param_name>=val` and `-F 'parameters={"hello": "world"}'` in the same request
    - File parameters still need to exist outside of the JSON string parameters if submitting through the command line
- Migrated API request parameter parsing to the Serializers

scott-taubman (changes requested):
Made some initial comments.  I'm mildly concerned about the amount of custom handling we're needing to do here, but it's at least contained to the view and serializer, so I'm inclined let us forge ahead with this if we feel sufficiently confident in how it works and the test coverage.  I'll probably come back for another round after thinking about it for a bit.

functionary/core/api/v1/utils.py
```diff
@@ -0,0 +1,114 @@
...
+from rest_framework import serializers
+from rest_framework.exceptions import ValidationError
+
+from core.models import Function, FunctionParameter
+from core.utils.parameter import PARAMETER_TYPE
+
+PREFIX = "param"
+SEPARATOR = r"\."
+
+r = re.compile(rf"^({PREFIX}){SEPARATOR}(\w+)$")
```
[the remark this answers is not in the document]
xeonneon: Fixed

functionary/core/tests/api/v1/views/test_task.py
```diff
@@ -32,31 +33,143 @@ def function(package):
...
+
+    def mock_file_upload(_task, _request):
+        """Mock the method of uploading a file to S3"""
+        return
+
+    url = reverse("task-list")
+
+    mocker.patch("core.api.v1.views.task._upload_files", mock_file_upload)
+
+    with open("core/tests/api/v1/views/test_text.txt", "rb") as f:
```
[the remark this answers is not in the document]
xeonneon: Fixed

functionary/core/api/v1/utils.py
```diff
@@ -0,0 +1,114 @@
...
+
+    Returns:
+        None
+
+    Raises:
+        ValidationError: When a JSON parameter is not valid JSON
+    """
+    function = _get_function(values)
+
+    json_params: list[FunctionParameter] = [
```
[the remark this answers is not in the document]
xeonneon: Fixed

functionary/core/api/v1/utils.py
```diff
@@ -0,0 +1,114 @@
...
+            values["parameters"][json_param.name] = json.loads(
+                values["parameters"][json_param.name]
+            )
+        except json.decoder.JSONDecodeError as err:
+            exception_map = {json_param.name: err.msg}
+            exc = ValidationError(exception_map)
+            raise serializers.ValidationError(serializers.as_serializer_error(exc))
+
+
+def _get_function(values: OrderedDict) -> Function:
```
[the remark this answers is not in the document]
xeonneon: Fixed. I opted to place the function validation after the `super().to_internal_value()` since I only wanted to access  the `function_name` and `package_name` fields once, and the `super().to_internal_value()` check expects those those two fields to exist. So I figured I could just pop those fields after the `super().to_internal_value()` check

functionary/core/api/v1/views/task.py
```diff
@@ -52,7 +61,15 @@ def get_serializer_class(self):
...
             "Execute a function. The function to be executed can be defined either "
             "by supplying function as a string uuid, or function_name and "
-            "package_name."
+            "package_name. "
+            "Parameters can be passed to the API either by prefixing them with "
+            f"`{RENDER_PREFIX}`, or placing them inside a `parameters` JSON "
+            "string. Prefixed parameters will **override** duplicate parameters "
+            "inside the parameters JSON string. "
+            "An example usage of parameters is as follows: "
+            "`-F 'param.file=@/path/to/README.md'` "
```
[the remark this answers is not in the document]
xeonneon: Removed the curl specific bits

functionary/core/api/v1/views/task.py
```diff
@@ -67,16 +80,26 @@ def get_serializer_class(self):
...
+
         request_serializer.is_valid(raise_exception=True)
 
         request_serializer.save(
-            creator=self.request.user,
-            environment=self.get_environment(),
+            creator=self.request.user, environment=self.get_environment()
         )
 
+        _handle_file_parameters(request, request_serializer)
```
scott-taubman, line 99: There's a race condition here because the task is already saved when you get to this point, so if this file upload doesn't happen fast enough the tasking will have run already without the file being there.

I think there's a couple things may want to do or thing about doing here.

1) We should probably separate the task publication (the thing that's actually going to result in the tasking reaching the runner) from the task instance creation.  Right now the publish happens automatically in the `post_create`.  If we instead make it an explicit step, where you have to call `task.publish()`, then you can control when that happens and make sure you only call it after the file parameters have been handled.

2) Given the above, we'll probably need some error handling for after you save the task but before you publish in the event that the file upload fails.  Not just in terms of the API response error, but marking the task itself as having errored.  This is kind of not ideal though because it's not really the tasking that failed (the function didn't run and fail) it was submitting the task that failed.  It seems like the ideal behavior would be to have never actually created the task if the file upload is unsuccessful, but we need some info from the task (sort-of) in order to do the file handling.  The easiest solution to that is to wrap the task creation and file upload stuff in a transaction.  There are some risks there given that the file upload could take a long-ish time in some circumstances, but it's probably the thing to do for now.

Maybe do the transaction thing now and then if you want to put a `#TODO` comment in to come back and address the race condition because it has implications in other parts of the code, I would be okay with that.

scott-taubman: I had a belated realization that I should check how we're doing this in the UI view side of things, which revealed what is probably the right / easier solution to all of this for now....

You could probably just not call `request_serializer.save()` and instead just instantiate a `Task` using the `validated_data`.  Then call `task.save()` after you've done the file handling.  That's basically what happens on the UI side.

xeonneon: Went with the third recommendation. I think one follow on task from this issue could be to centralize the `Task` creation logic. I've spread the logic between both the API and the UI view, and they can be combined or maybe moved to a helper method for the Task model

functionary/core/api/v1/views/task.py
```diff
@@ -114,3 +137,32 @@ def log(self, request, pk=None):
...
+    """
+    if not request.FILES:
+        return
+
+    # Wrap items in list to avoid dictionary changed size error
+    for param, _ in list(request.FILES.items()):
+        if param_name := get_parameter_name(param):
+            request.FILES[param_name] = request.FILES.pop(param)[0]
+
+    task = Task.objects.get(id=request_serializer.instance.id)
```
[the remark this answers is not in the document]
xeonneon: Fixed

scott-taubman (approved):

Issue #201, opened 2023-02-06
Switch to a better supported CSS component library


xeonneon:
In an effort to move away from Bulma due to a lack of new features and extensibility, we should switch which CSS component library we use. We should migrate to a component library that allows for extensibility so the Functionary UI has some personality to it, and ideally comes with accessibility features, such as focus trapping, out of the box.

xeonneon:
I'm strongly in favor of a component library that is built on top of Tailwind CSS.

Some example component libraries include:
- [Tailwind Elements](https://tailwind-elements.com/)
- [DaisyUI](https://daisyui.com/)
  - Missing accessibility features
- [Tailwind UI Kit](https://tailwinduikit.com/components)
  - Missing accessibility features

scott-taubman:
Current design thoughts based on recent discussions:

* Just use the default bootstrap colors for now.  We'll figure out a better / more compliant color scheme later
* The [bootstrap sidebar example](https://getbootstrap.com/docs/5.0/examples/sidebars/) basically looks like what I believe we want.
* Let's get rid of the top bar now so that the side bar looks good and we don't have to worry about picking a good complimentary color for the top bar.

PR #216, opened 2023-02-14
#201 - swap out Bulma for Bootstrap

obr42:
Does not close ticket, only initial work.

Add in Bootstrap framework ~with hot reload on dev env only~ (will be added as opt-in on future branch). Begin conversion of files - base to include ~navbar and~ (removed) and navigation in side drawer (with a touch of refactoring). Main landing/welcome page ("welcome, user!") with alerts, non-admin social accounts page, login page and social account modal. 

Swap out icon png for font-based format. There is a bash script to pull out just the F letter that we want from the chosen font tff package and convert it into a woff2.

Made a `base_bulma.html` that is still just Bulma to point non-updated files to. 
Made a `modal_old.html` that is the now outdated modal template.

Note about side drawer - it will convert to just icons once the page size is small enough (based on px width). Right now it looks a bit weird because of the select box for environment. Went down a rabbit hole trying to make it better without a total refactor, but will have to wait until the other ticket to refactor page layout. It needs a more comprehensive refactor to get away from the select box, which isn't exactly in scope.

Add bash script to automatically compile CSS file from SASS files. It will download Bootstrap 5.2.3, use its .scss files to build from, and then remove all downloaded and related files. The `custom.scss` contains our overrides and imports the Bootstrap components we are currently using. There is some stubbed out imports that may need to be un-commented should we end up using those components.


kvig (changes requested):

functionary/ui/templates/base.html
```diff
@@ -45,124 +48,97 @@
...
+                                <a class="nav-link px-0" href="{% url 'ui:scheduledtask-list' %}">
+                                    <i class="fa fa-lg fa-clock text-secondary"></i>
+                                    <span class="ms-1 d-none d-md-inline">Schedules</span>
+                                </a>
+                                {% if user.is_superuser %}
+                                    <a class="nav-link px-0" href="{% url 'ui:environment-list' %}">
+                                        <i class="fa fa-lg fa-building text-secondary"></i>
+                                        <span class="ms-1 d-none d-md-inline">Environments</span>
+                                    </a>
+                                    <a class="nav-link px-0" href="{% url 'ui:team-list' %}">
```
scott-taubman, line 102: Small thing that I note here since this is where it's most noticeable...

The alignment of the menu text is off.  With the previous styling, the icons were all centered with each other and then there was consistent spacing so that the text was also all aligned.  Now the icons are all left-aligned with the text not really aligned at all.  The bootstrap [sidebars example](https://getbootstrap.com/docs/5.0/examples/sidebars/#) shows alignment similar to what we had, so it seems like it should be doable.

functionary/ui/templates/partials/modal.html
```diff
@@ -1,34 +1,37 @@
...
-        </section>
-        {% block modal_nofooter %}
-            <footer class="modal-card-foot">
-                {% block modal_footer %}
-                {% endblock modal_footer %}
-            </footer>
-        {% endblock modal_nofooter %}
-        {% block form_end %}
-        {% endblock form_end %}
+<div class="modal" id="form-modal">
```
scott-taubman, line 1: Changing the modal now breaks all of the pages still on bulma that use modals (which is most of them).  You probably need to create a `modal_bootstrap.html` partial and point specific things at that as you update them (like the login page) while leaving the pages that are still on bulma pointing at the existing `modal.html`.

functionary/ui/templates/registration/login.html
```diff
@@ -1,34 +1,29 @@
-{% extends "base.html" %}
-
+{% extends "base_old.html" %}
```
scott-taubman, line 1: Can we call this `base_bulma.html`?  Also, inverting it and creating a `base_bootstrap.html` and then pointing pages at that as you convert them would have made for a smaller initial change set here, since pages that you're not intending to change would be untouched for now...

functionary/ui/scss/custom.scss
```diff
@@ -0,0 +1,78 @@
...
+@import "./functions";
+@import "./variables";
+@import "./mixins";
+@import "./maps";
+
+// 2. Include any default variable overrides here
+
+$white: #FFFFFF;
+$black: #000000;
+$blue: #485FC7;
```
scott-taubman, line 14: As mentioned elsewhere, it feels like we shouldn't be redefining these colors.

functionary/ui/templates/registration/login.html
scott-taubman: I don't feel that strongly about it.  You can just do the rename to `base_bulma.html` and then carry on as you were.

functionary/ui/templates/account/login.html
```diff
@@ -37,105 +33,86 @@
...
                                                                     hx-get="{% provider_login_url provider_pair.provider.id openid=brand.openid_url process=process %}">
-                                                                &nbsp;{{ brand.name }}
+                                                                {{ brand.name }}
                                                             </button>
                                                         </div>
                                                     {% endfor %}
                                                 {% else %}
-                                                    <div class="column">
+                                                    <div class="col">
+                                                        {% csrf_token %}
```
kvig, line 108: Just curious, is this actually sending the CSRF token with the login with <account> request?

functionary/ui/templates/base.html
```diff
@@ -45,124 +41,93 @@
...
+                                    <i class="fa fa-lg fa-user fa-fw"></i>
+                                    <span class="ms-1 d-none d-md-inline">{{ user }}</span>
+                                </a>
                                 <form method="post" action="{% url 'account_logout' %}">
                                     {% csrf_token %}
-                                    <button class="button is-underlined is-ghost has-text-white pr-0 pb-0"
-                                            type="submit">
-                                        <span class="icon has-text-danger mr-2"><i class="fa fa-door-open"></i></span>Logout
+                                    <button class="btn nav-link px-0" type="submit">
+                                        <i class="fa fa-md fa-door-open fa-fw text-primary"></i>
```
scott-taubman, line 106: This icon is smaller than the others, which makes it not line up nicely with all of the other menu items.

functionary/functionary/settings/debug.py
```diff
@@ -4,10 +4,13 @@
 
 DEBUG = True
 
-INSTALLED_APPS += ["debug_toolbar"]
+INSTALLED_APPS += ["debug_toolbar", "django_browser_reload"]
```
scott-taubman, line 7: There's an issue either with this `django_browser_reload` app and middleware or our usage of it.  Basically every time you load a page in the UI (either by navigating or because of the new autoreload) a new database connection is opened.  The result is that just using functionary normally eventually exhausts the available connections to the database and the whole thing stops working.

We either need to figure out what's going on there and how to fix it, or remove this app.

kvig: What did you use to get the connection count?

functionary/ui/scss/custom.scss
```diff
@@ -0,0 +1,132 @@
...
+// 2. Include any default variable overrides here
+$white: #FFFFFF;
+$black: #000000;
+$gunmetal: #343E40;
+
+$primary: #A05F64;
+$secondary: #C6C6C3;
+$panel: #D4D4D2;
+$body-bg: #F7F8F8;
+$tertiary: #06647F;
```
xeonneon, line 16: Thoughts on switching the Functionary header panel to the tertiary color? If that breaks accessibility, then I guess it's fine. I think the tertiary's blue is a bit more welcoming than the red

xeonneon: I also prefer the tertiary color as the primary color, which was the original reason I had included this section of the scss

functionary/ui/templates/base.html
```diff
@@ -32,6 +23,11 @@
         <link rel="stylesheet"
               href="{% static 'css/functionary.css' %}"
               type="text/css"/>
+        {# Load the bootstrap libraries #}
+        {% bootstrap_css %}
+        {% bootstrap_javascript %}
```
xeonneon, line 28: This ends up pulling the bootstrap JS and css files from a CDN. I think for the purpose of version control, we might want to just download the bootstrap files ourselves and load it like the rest of our static assets

functionary/functionary/settings/debug.py
scott-taubman: > What did you use to get the connection count?

```sql
select count(*) from pg_stat_activity;
```

If you run that from a psql shell you'll notice that it goes up by 1 on every page load.  If you take `django_browser_reload` out of the picture that does not happen.

If this was an inherent problem with the library it would be unusable so I have to imagine it's some weird interaction between it and something else we're doing.  I haven't spent much time debugging it beyond noting described behavior, but we cannot have it enabled by default for our dev setup without figuring this out.

If we can't figure it out, then we can do the work to make it opt-in in the dev-setup scenario so that if some devs want to put up with the drawbacks they can, but not everyone is forced to.

kvig: I'm not seeing this with either the middleware or the script tag in the home.html. I do briefly see an extra connection when making .py changes but the connection goes away in a reasonably short amount of time. Making changes to the .html shows up automatically as expected without extra connections.

scott-taubman (approved):

PR #228, opened 2023-02-23
Import local.py for additional settings if present

kvig:
This PR updates the `debug.py` settings to attempt to import from `local_settings.py`if available. Out of the box, this will add all new entries to the `django.conf.settings` with whatever values are in the `local_settings.py`. Any matching values will be overridden.

### DIFF
--- a/functionary/functionary/settings/debug.py
+++ b/functionary/functionary/settings/debug.py
@@ -17,7 +17,4 @@ INTERNAL_IPS = [ip[: ip.rfind(".")] + ".1" for ip in ips] + [
     "10.0.2.2",
 ]
 
-try:
-    from .local import *
-except ImportError:
-    pass
+from .local_settings import *  # noqa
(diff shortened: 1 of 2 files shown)


functionary/functionary/settings/debug.py
```diff
@@ -16,3 +16,8 @@
     "127.0.0.1",
     "10.0.2.2",
 ]
+
+try:
+    from .local import *
```
scott-taubman, line 21: I think you can solve the problem you were having by just importing the base stuff in local.py, and then putting this import immediately after the import from base here so:

```python
# local.py
from .base import *

INSTALLED_APPS += ["django_browser_reload"]
SOMESETTING = "HEY"
```

```python
# debug.py
import socket

from .base import *  # noqa
from .local import *  # noqa

# other stuff as usual
```

The fact that both local.py and debug.py import from base shouldn't cause a problem.  Also, I'd say just check in a local.py that has the base import and nothing else (maybe it has a comment explaining what it's for).  Then you don't need the try / except here and can instead just count on the file always being there.

And finally, add the local.py to the .gitignore so that people aren't accidentally checking in their personal settings.

kvig: The .gitignore already had local_settings.py included so I moved it to that filename.

I left the local_settings import last so it can override all settings.

PR #229, opened 2023-02-23
#224 - Fix optional parameter handling when tasking through the UI

### LINKED ISSUE #224: Optional function parameters are not properly handled
There are currently several problem related to optional parameter handling in the UI views, a lot of which has to do with the fact that optional parameters are always included, but get set to `None`, `''`, `False` or some other default value.  That leads to the following behaviors:

* The `validate_parameters` check will always fail for certain data types if an optional parameter is not included.  Numbers and dates for instance will fail to validate because a value of `None` is supplied, which is not a valid value for those types.

* If we bypass that that validation, any optional parameters that were not specified still saved as if they were supplied, but with some default empty value.  For the given types, they are stored as:
  ```python
  {
    'boolean_': False,
    'date_': None,
    'datetime_': None,
    'file_': '',
    'float_': None,
    'integer_': None,
    'json_': None,
    'string_': '',
    'text_': ''
  }
  ```
  This leads to lots of problems, such as the inability in the function to distinguish between an optional parameter that was not supplied, versus one that was supplied as an empty string, for instance.

* For files, we still generate the signed download URL, even though no actual file was ever uploaded to that location.  So not only does it look like a file was supplied even if one wasn't, a URL to nowhere gets supplied to the function.

scott-taubman:
Closes #224 

This PR fixes optional parameter handling so that fields left blank on the tasking form are not saved as part of the Task parameters.  The prior behavior resulted in every parameter being included, with those that were blank on the form getting set to the Field class's `empty_value`, which is either `None` or `""` depending on the field type.

Note: boolean fields will always be included.  The empty value for a boolean is `False`, which makes sense, since that field being left blank on the form implies a value of `False`.

## Testing
A unit test has been added to ensure the proper behavior.  This can be easily manually tested by running the "long running process" example function and trying to execute with no value set for duration.  Previously this would result in a validation error.  Now it is accepted and the Task detail page will show the parameters as `{}`.

### DIFF
--- a/functionary/ui/tests/views/test_function.py
+++ b/functionary/ui/tests/views/test_function.py
@@ -1,0 +1,65 @@
+"""Tests function views"""
+
+import pytest
+from django.urls import reverse
+
+from core.models import Function, Package, Task, Team
+from core.utils.parameter import PARAMETER_TYPE
+
+
+@pytest.fixture
+def environment():
+    team = Team.objects.create(name="team")
+    return team.environments.get()
+
+
+@pytest.fixture
+def package(environment):
+    return Package.objects.create(name="testpackage", environment=environment)
+
+
+@pytest.fixture
+def function(package):
+    return Function.objects.create(
+        name="testfunction",
+        package=package,
+        environment=package.environment,
+    )
+
+
+@pytest.fixture
+def integer_parameter(function):
+    return function.parameters.create(
+        name="optional_integer", parameter_type=PARAMETER_TYPE.INTEGER, required=False
+    )
+
+
+@pytest.fixture
+def string_parameter(function):
+    return function.parameters.create(
+        name="optional_string", parameter_type=PARAMETER_TYPE.STRING, required=False
+    )
+
+
+@pytest.mark.django_db
+def test_execute_handles_optional_parameters(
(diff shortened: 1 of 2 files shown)


xeonneon:
Works as advertised. Although I do wonder if it's worth to include the default value in the parameters if the parameter is optional. For example, in the `long_running_process`, I only know what that default value is if I go to the Function detail page. If I look at the task details, I don't see any parameters, so it might not be intuitive for user's to recognize that the default values were used to run the task.

functionary/ui/tests/views/test_function.py
```diff
@@ -0,0 +1,65 @@
...
+        "function_id": str(function.id),
+        f"task-parameter-{integer_parameter.name}": "",
+        f"task-parameter-{string_parameter.name}": "",
+    }
+
+    admin_client.post(url, data)
+    task = Task.objects.get(function=function)
+
+    assert integer_parameter.name not in task.parameters
+    assert string_parameter.name not in task.parameters
```
kvig, line 65: This should pass in a populated parameter to make sure it doesn't remove all of the parameters.

scott-taubman: Good call.  I'll sneak that in to whatever I commit next.

PR #230, opened 2023-02-24
#201 - task detail

obr42:

Converts task detail page into bootstrap. A few tweaks to base.html to make sidebar not scroll when main section needs to scroll. Removes nested container from login and social account page (bad structure).

Wraps contents of task result to prevent horizontal page scroll as much as possible. Uses pagination classes on the toggle buttons for result format to create consistency.


functionary/ui/templates/partials/task_result_block.html
```diff
@@ -1,22 +1,28 @@
...
-        </div>
-        {% include "partials/task_result.html" %}
+<div id="result-block" hx-swap-oob="true">
+    <div class="ps-4">
+        {% if completed %}
+            <div class="py-3" hx-target="#result-block">
+                <div aria-label="Result output type">
+                    <ul class="pagination">
+                        <li class="page-item {% if output_format != "table" and output_format != "log" %}active{% endif %}">
+                            <a hx-get="{% url 'ui:task-results' task.id %}?output=display_raw"
```
scott-taubman, line 8: There's a slightly weird behavior here where the cursor doesn't update to a pointer like you'd expect.  This seems to be because we have an <a> with no `href`, so the browser pointer for links doesn't get applied.  And none of the classes we're using explicitly change the pointer.  Two options come to mind and I'm sure there are more:

* Make these <button> rather than <a>.  That seems to maintain the look, but get us the expected cursor.  It may be more accurate too since I guess they are behaving more like buttons than links.
* Add an `href="#"` or something similar so that the default browser styling is applied.

I don't know which of those seems more appropriate.  There may be other options too.

PR #231, opened 2023-02-27
#146 Added filter for list views

### LINKED ISSUE #146: Add filtering support to the list views
Users should have the ability to search for particular functions or packages, rather than needing to scroll through a paginated list. Users should also have the ability to apply filters to the different list views. The filters could be datetime fields (created before/after some date), the tasks that used a particular function, the tasks executed by a particular user, etc.

west270:
Closes #146

Added filters to list views using https://github.com/carltongibson/django-filter.


functionary/ui/templates/partials/table.html
```diff
@@ -1,2 +1,25 @@
 {% load django_tables2 %}
-{% render_table table 'django_tables2/bootstrap5.html' %}
+{% load widget_tweaks %}
+{% if filter %}
+    <form action="" method="get" class="form row">
```
scott-taubman, line 4: Let's put this stuff in its own partial (templates/partials/filter.html).  Then here you would just include that.

functionary/ui/tables/function.py
```diff
@@ -1,9 +1,20 @@
+import django_filters
 import django_tables2 as tables
 from django.urls import reverse
 
 from core.models import Function
 from ui.tables.meta import BaseMeta
 
+FIELDS = ("name", "package", "summary")
```
scott-taubman, line 8: I would leave summary out of the filterable fields for now.

functionary/ui/tables/build.py
```diff
@@ -1,10 +1,33 @@
+import django_filters
 import django_tables2 as tables
 from django.urls import reverse
 
 from builder.models import Build
 from ui.tables import DATETIME_FORMAT
+from ui.tables.filters import DateTimeFilter
 from ui.tables.meta import BaseMeta
 
+FIELDS = ("package__name", "status", "created_at", "creator")
```
scott-taubman, line 10: As is the label for package name ends up as "Package Name" rather than just "Package."  Which then lead me to notice that the table column is called "Name" rather than "Package".  We should fix both of those things.

functionary/ui/templates/partials/table.html
```diff
@@ -1,2 +1,25 @@
...
+        {% for field in filter.form %}
+            <div class="col-auto mb-3 ">
+                {{ field|add_label_class:"form-label" }}
```
scott-taubman, line 7: Small thing, but can we get rid of the ":" in the label?  Right now it's something like "Function:", but the ":" is kind of superfluous.

scott-taubman (commented):
Made some initial comments for now.  I think the styling needs some love still, but that's true of the tables as well, so it may make sense to work those things out together as a follow-on.

PR #232, opened 2023-02-27
Date / Datetime parameter validation

### LINKED ISSUE #225: Date parameters cannot be validated
There is currently no conversion of date and datetime input from the UI before the pydantic schema validation runs.  This results in the `jsonschema.validate` call being made against input that contains `date` and `datetime` types rather than some json valid string input.

If we are to continue using pydantic for parameter validation at the model level, we'll need to update the `validate_parameters` function to include a call to something similar to the current `_serialize_json_parameters` call that serializes `date` and `datetime` fields into an acceptable string format.  We also need to ensure that the `date` and `datetime` fields get serialized consistently when stored / sent out for tasking.

scott-taubman:
Closes #225 

This PR updates the Task parameter validation to serialize date and datetime fields in addition to JSON to allow them to properly validate against the json schema.  The result is that functions using date or datetime parameters can now be executed via the UI.

## Test Instructions

* New unit tests have been added to confirm that the `validate_parameters` function works for date and datetime parameters, when provided as either native python types or as properly formatted strings (improperly formatted values are covered by existing tests).
* The `parameter_types` demo function can now be tasked via the UI with values supplied for the `date_` and `datetime_` fields.  NOTE: They are received by the function as strings. This behavior is up for discussion, but is likely correct.


functionary/core/utils/parameter.py
```diff
@@ -99,13 +149,13 @@ def validate_parameters(parameters: dict, instance: Union["Function", "Workflow"
...
         pydantic_model = _get_pydantic_model(instance)
-        serialized_parameters = _serialize_json_parameters(parameters, instance)
+        serialized_parameters = _serialize_parameters(parameters, instance)
         jsonschema.validate(
             serialized_parameters, pydantic_model(**serialized_parameters).schema()
         )
     except (
         ValidationError,
-        jsonschema.exceptions.ValidationError,
+        jsonschema.ValidationError,
```
scott-taubman, line 158: Unrelated change to fix the linter yelling at me.

functionary/ui/tests/views/test_function.py
```diff
@@ -41,9 +41,16 @@ def string_parameter(function):
     )
 
 
+@pytest.fixture
```
scott-taubman, line 44: Sneaking in a suggested improvement to this unrelated test.

kvig (approved):
LGTM

PR #233, opened 2023-02-28
#223 - Fix pydantic BaseModel property shadowing in parameter validation

### LINKED ISSUE #223: Function parameters cannot use pydantic BaseModel names
Giving a function parameter a name that matches one of the pydantic BaseModel attributes results in an error on anything that attempts to access `function.schema`.  The exception thrown is:

`Field name "json" shadows a BaseModel attribute; use a different field name with "alias='json'".`

To reproduce this, create a function with a parameter named `json`, then try accessing that function in the UI.

scott-taubman:
Closes #223 

This PR fixes the issue of function parameter names not being able to validate if they shadow properties of pydantic's BaseModel (i.e. "json").

The actual fix is a one liner in [functionary/core/utils/parameter.py](https://github.com/beer-garden/functionary/compare/main...scott-taubman:223-pydantic-basemodel-collision-fix?expand=1#diff-c454bb02aef11ce5ab31c885aa281fc1409da80c936fe968eab00a4fad2695ea).  The rest is updates to the example function and some of the unit tests.

## Test Instructions
* The validation unit tests were updated so that the `json_param` fixture now uses one of the problematic names: "json". The one change would make all of the tests fail before this fix.  With the change, they continue to pass.
* The `parameter_types` demo function no longer appends `_` to the parameter names, which was done previously as a workaround to this specific problem.  Publishing and tasking this updated version should work fine rather than yelling at you about the parameter named "json".

### DIFF
--- a/examples/demo/functions.py
+++ b/examples/demo/functions.py
@@ -1,6 +1,5 @@
 import os
 import time
-from datetime import date, datetime
 
 import requests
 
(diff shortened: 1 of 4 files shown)


kvig (approved):
Worked as I expected

PR #237, opened 2023-03-01
#201 - build detail

obr42:

Converts Build detail page into bootstrap. Edit base.html to start handling page content width better and more fully utilizing Bootstrap's responsive breakpoints. Responsive width is [preferred](https://ux.stackexchange.com/questions/28124/what-is-the-current-pixel-width-standard-for-a-websites-content-area#:~:text=The%20Right%20Answer%20is%20Responsive%20Design) approach [over](https://ux.stackexchange.com/questions/24445/optimal-fixed-width-website-design#:~:text=If%20you%20serve%20a%20very%20large%20amount%20of%20users%20and%20there%20is%20no%20clear%20cut/primary%20screen%20size%20(view%20port%20for%20a%20more%20modern%20approach)%2C%20I%20would%20suggest%20looking%20into%20other%20avenue%20like%20Responsive%20Design.) fixed-width. Extra [whitespace added](https://www.thoughtco.com/web-page-widths-3469968#:~:text=If%20the%20scan%20length%20of%20a%20line%20of%20text%20is%20longer%20than%2010%20to%2012%20words%20or%20shorter%20than%204%20to%205%20words%2C%20it%20can%20be%20difficult%20to%20read.) to manage [main content column width](https://socialtriggers.com/perfect-content-width/#:~:text=ranges%20between%2040%20and%2055%20characters%20per%20line%2C%20or%20in%20other%20words%2C%20a%20content%20column%20that%20varies%20between%20250%2D350%20pixels%20wide), to ensure users are [not overwhelmed](https://www.lemonandthesea.com/blog/the-best-width-for-designing-your-website#:~:text=The%20ideal%20width%20of%20a%20column%20is%20around%2012%2D17%20words%20per%20line) with information.

Add checkbox toggle to wrap (or not) log text. When wrapped, lines break and are visible without horizontal scroll. When not wrapped, lines do not break and the `<pre>` adds its own content horizontal scrollbar (i.e. page does not scroll, just the log content). Users can scroll horizontally by clicking on or highlighting text of the log, then using left/right arrow keys to move left/right. Toggle is a CSS only solution, but does use a selector that is [not enabled by default](https://developer.mozilla.org/en-US/docs/Web/CSS/:has#browser_compatibility ) in Firefox (but is supported, users just may need to enable this feature). Disabling text wrap does not remove the general content boundaries (i.e. right-sided white-space) that are creating a more user-digestible presentation.


functionary/ui/static/css/functionary.css
```diff
@@ -29,6 +29,11 @@
 .errors {
     color: red;
 }
+/* the :has() must be user enabled to work on firefox */
+/* https://developer.mozilla.org/en-US/docs/Web/CSS/:has#browser_compatibility */
+div:has(input[type="checkbox"][name="wrap"]:not(:checked)) ~ pre {
```
scott-taubman, line 34: This toggle doesn't seem to work in firefox for me even after enabling the feature. 

image

image

Not a show stopper, but it seemed like maybe this isn't working as intended?

functionary/ui/templates/builder/build_detail.html
```diff
@@ -1,58 +1,56 @@
...
-                {% if build_log.log %}
-                    <div class="field">
-                        <label class="label" for="desc">Build Log:</label>
-                        <pre class="ml-4">{{ build_log.log }}</pre>
+            {% endif %}
+            {% if build_log.log %}
+                <div>
+                    <strong class="fs-4"><i class="fa fa-book fa-sm fa-fw pe-1"></i>Build Log:</strong>
+                    <div class="ps-4">
+                        <div class="form-check py-3">
```
scott-taubman, line 42: Can we add `form-switch` here to make it a switch instead of a checkbox?  Seems like a better fit.

functionary/ui/static/css/functionary.css
scott-taubman: That bug you link to was opened 15 years ago, so despite some recent activity on it, I won't hold my breath for it to be fixed soon.

It doesn't really hurt anything to have it in there for now so I'm inclined to just merge this.  We can circle back and try other options that have better cross-browser compatibility at a later time.

PR #239, opened 2023-03-01
Fix environment selector rendering

scott-taubman:
This PR updates the handling of the environment selector drop down so that it doesn't flash in and out as it's re-rendering on every page load.

What I've done is moved the logic that was in `EnvironmentSelectView` into a context processor so that the user environment data gets stamped into the context automatically (whenever `get_context_data()` is called).  That lets us get out of having to do the lazy load of the environment selector and can instead just render it directly as part of the main template rendering.

## Test Instructions
* Note the current behavior of the env selector dropdown flashing in and out as you change pages.
* On this branch that should no longer be the case.


kvig (approved):
Works for me

PR #240, opened 2023-03-01
#201 - swap package detail

obr42:

Converts Package detail page into bootstrap. Globally removes underline from links - cleanup related code. Also removes `:` from content headings.


functionary/ui/templates/core/package_detail.html
```diff
@@ -1,52 +1,40 @@
...
+                    <a class="text-decoration-none" href="{% url 'ui:package-list' %}">Package List</a>
+                </li>
+                <li class="breadcrumb-item active" aria-current="page">{{ package.render_name }}</li>
+            </ol>
+        </nav>
+        <h1 class="py-2">
+            <i class="fa fa-cubes fa-xl"></i>
+            {{ package.render_name }}
+        </h1>
+        <div class="vstack gap-4 py-2">
```
scott-taubman, line 16: Something about the spacing here feels off.  The Summary  or Description feels too close to the package name and then the space between Summary and Functions is a little too large.  Playing around with it, removing the `gap-4` here and changing the h1 to be `py-4` looks more evenly spaced, but play around with it.

functionary/ui/templates/core/package_detail.html
```diff
@@ -1,52 +1,40 @@
...
+            {% if package.description %}
+                <div>
+                    <strong class="fs-4"><i class="fa fa-newspaper fa-sm fa-fw pe-1"></i>Description:</strong>
+                    <p class="ps-4 py-2">{{ package.description }}</p>
+                </div>
+            {% elif package.summary %}
+                <div>
+                    <strong class="fs-4"><i class="fa fa-rectangle-list fa-sm fa-fw pe-1"></i>Summary:</strong>
```
scott-taubman, line 24: This isn't related to a thing you changed, but since you're here....Can we remove the ":" from these headings?  And just do that everywhere going forward?  They don't really seem necessary.

functionary/ui/templates/core/package_detail.html
```diff
@@ -1,52 +1,40 @@
...
+                <strong class="fs-4"><i class="fa fa-cube fa-sm fa-fw pe-1"></i>Functions:</strong>
+                <div class="ps-4">
+                    <div id="funcs" class="list-group">
                         {% for function in functions %}
-                            <li class="block ml-4">
-                                <a href="{% url 'ui:function-detail' function.id %}">{{ function.render_name }}</a>
-                                <br/>
-                                <span class="ml-4">{{ function.summary }}</span>
-                            </li>
+                            <a href="{% url 'ui:function-detail' function.id %}"
```
scott-taubman, line 33: May or may not be related to this change, but I'll ask the question here:  Is there ever a case where we'll want the underlines on links?  In almost all cases it seems to look better without them.  I ask that here because depending on the answer we could either:

* Just turn them off universally, which then isn't necessarily part of this PR (though it could be).
* Turn them off specifically for these links, in which case it is definitely relevant to this PR.

PR #243, opened 2023-03-02
#241 - Added failed file upload error handling

### LINKED ISSUE #241: Add exception handling on failed file upload

xeonneon:
Closes #241 

## Introduced Changes
- Added a `S3FileUploadError` exception
  - Raised when `client.put_file` raises any exception
- API and UI return an error claiming that the file(s) were not able to be uploaded with a status of `503`

## Testing
- Added unit tests for API and views
- To manually test, simply raise the `S3FileUploadError` before the `client.put_file` in `minio.py` to view the errors that get returned or rendered

### DIFF
--- a/functionary/core/api/v1/views/task.py
+++ b/functionary/core/api/v1/views/task.py
@@ -27,7 +27,7 @@ from core.api.v1.serializers import (
 from core.api.v1.utils import PREFIX, SEPARATOR, get_parameter_name
 from core.api.viewsets import EnvironmentGenericViewSet
 from core.models import Task, TaskResult
-from core.utils.minio import S3FileUploadError, handle_file_parameters
+from core.utils.minio import S3Error, handle_file_parameters
 
 RENDER_PREFIX = f"{PREFIX}{SEPARATOR}".replace("\\", "")
 
@@ -106,7 +106,7 @@ class TaskViewSet(
             task.save()
         except (ValidationError, DjangoValidationError) as err:
             raise serializers.ValidationError(serializers.as_serializer_error(err))
-        except S3FileUploadError as err:
+        except S3Error as err:
             return Response(
                 {"error": f"{err}"},
                 status=status.HTTP_503_SERVICE_UNAVAILABLE,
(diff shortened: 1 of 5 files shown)


functionary/core/tests/api/v1/views/test_task.py
```diff
@@ -228,6 +230,36 @@ def mock_file_upload(_task, _request):
...
+        data=task_input,
+        content_type=MULTIPART_CONTENT,
+        **request_headers,
+    )
+
+    task_id = response.data.get("id")
+
+    assert response.status_code == status.HTTP_503_SERVICE_UNAVAILABLE
+    assert task_id is None
+    assert not Task.objects.filter(id=task_id).exists()
```
scott-taubman, line 260: This assertion seems superfluous given the previous one.  If `assert task_id is None` passes, this is guaranteed to be false.  In the similar test for the ui view you filter on the `file_function` which seems like a good alternative.

functionary/core/api/v1/views/task.py
```diff
@@ -106,6 +106,11 @@ def create(self, request: Request, *args, **kwargs):
             task.save()
         except (ValidationError, DjangoValidationError) as err:
             raise serializers.ValidationError(serializers.as_serializer_error(err))
+        except S3FileUploadError as err:
```
scott-taubman, line 109: A couple of things here:

* If S3 is not reachable at all, an S3ConnectionError is what gets raised, which isn't covered here.  You can add that here, or probably even better would be to go back and make S3FileUploadError and S3ConnectionError both build off of a base exception called S3Error and handle that here.  That way we have a catch-all for these situations where we want to handle any S3 related error the same way and don't have to call them all out individually (even though "all" is only 2 right now).

* The error message returned here should really be the same one that we show in the UI.  It's an API user facing error, not a backend administrator facing one.  So showing the API user something like `Error: HTTPConnectionPool(host='localhost', port=9000): Max retries exceeded with url...` is not really appropriate.

xeonneon: Added parent `S3Error` exception class that `S3ConnectionError` and `S3FileUploadError` extend from, so now the view and API only need to catch the generic `S3Error`. Also updated the error message that those exceptions store such that they contain sensible user-facing error messages, while the logger messages include the actual exception error for administrators.

PR #244, opened 2023-03-03
Adds page for general Constance settings.

kvig:
This PR renames the `core` entry in the Django Admin to `Functionary`. It also adds a settings page for general settings stored in Constance. There is a `HIDDEN` entry in the fieldset to hide any settings that should not be changed from that page. Any settings not added to the `CONSTANCE_CONFIG_FIELDSETS` value will be complained about on startup and should either be added to a new fieldset, or to the `HIDDEN` fieldset.

If using `docker compose` to start Functionary, you'll need to set the host to `minio`, the other settings should be sufficient.

Testing:
Try to run the `Number of Characters` function as is. It should fail since the connection information is incorrect. Change the S3 host to `minio` in the DJango Admin and restart the containers. The function should now complete when executed.


functionary/ui/templates/admin/constance/change_list.html
```diff
@@ -0,0 +1,43 @@
+{% extends "admin/constance/change_list.html" %}
```
scott-taubman, line 1: Is this extending itself?

functionary/functionary/settings/base/constance_.py
```diff
@@ -0,0 +1,67 @@
...
+        },
+        "SocialAccount Providers Configuration",
+        "config_field",
+    ),
+    "S3_HOST": ("localhost", "Server Hostname", str),
+    "S3_PORT": (9000, "Port", int),
+    "S3_REGION": ("", "Region", str),
+    "S3_SECURE": (False, "Require Secure Access", bool),
+    "S3_ACCESS_KEY": ("ACCESS_KEY", "ROOTUSER", str),
+    "S3_SECRET_KEY": ("SECRET_KEY", "CHANGEME123", str),
```
scott-taubman, line 21: The order of these settings appears to be (default value, name, type), meaning "CHANGEME123" is what is being shown as the setting name in the admin pages.  I'm assuming that's not the intention.

functionary/ui/templates/admin/constance/includes/results_list.html
```diff
@@ -0,0 +1,63 @@
...
+                <td>{{ item.default|linebreaks }}</td>
+                <td>
+                    {{ item.form_field.errors }}
+                    {% if item.is_file %}
+                        {% trans "Current file" %}: <a href="{% get_media_prefix as MEDIA_URL %}{{ MEDIA_URL }}{{ item.value }}"
+    target="_blank">{{ item.value }}</a>
+                    {% endif %}
+                    {{ item.form_field }}
+                    {% if not item.is_file %}
+                        <br/>
```
xeonneon, line 43: Just a note of some css fix to come back to: The padding on the <p> elements in this particular table cell causes an out-of-place indentation. IMO it would look better if there was no indentation, so that it aligns with the left-alignment like the other cells

functionary/functionary/settings/base/constance_.py
```diff
@@ -0,0 +1,67 @@
...
+    "SOCIALACCOUNT_PROVIDERS": (
+        {
+            "github": {},
+            "gitlab": {},
+            "keycloak": {},
+        },
+        "SocialAccount Providers Configuration",
+        "config_field",
+    ),
+    "S3_HOST": ("localhost", "Server Hostname", str),
```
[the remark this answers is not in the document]
kvig: Changed the strings to empty string and added the defaults from the `docker-compose.yml` to the `bootstrap.yaml` for first load. I also updated `minio.py` to check for the `S3_HOST` in the environment.

scott-taubman (approved):

PR #246, opened 2023-03-03
#201 - function detail

obr42:

Converts Function detail page into bootstrap. Removes extra font import.


functionary/ui/templates/forms/task_parameters.html
```diff
@@ -1,13 +1,11 @@
...
-    <div class="control">
-        {{ field }}
+    <div class="mb-4">
+        <label class="form-label" for="{{ field.id_for_label }}">
+            <b {% if field.errors %}class="text-danger"{% endif %}>{{ field.label }}</b>:
+            <span class="text-muted">{{ field.field.label_suffix }}</span>
+        </label>
+        <div>{{ field }}</div>
+        <div id="{{ field.label }}HelpBlock" class="form-text">{{ field.help_text }}</div>
+        {{ field.errors }}
```
scott-taubman, line 9: The rendering of this brought to attention that some of the stuff in `functionary.css` needs to get updated.  Specifically, there are hardcoded references to "red" that ought to be `var(--bs-danger)`.  Since that's not really related to this change, I'm going to merge this and we can update the css file in a future PR.

scott-taubman (approved):

PR #247, opened 2023-03-03
#206 - Added builder exception handling

### LINKED ISSUE #206: Builder should record errors on any exceptions thrown
There are some types of errors that can get thrown during the package build process that do not result in the build status getting marked as ERROR and a build log being saved.  Instead, we end up with the build stuck IN_PROGRESS with no useful information presented to the user.

Basically anything in the `build_package` function that is not wrapped in a try / except block is subject to this problem.  We should ensure that the entire code path has exception handling and that the build status and log get updated appropriately when an error occurs.

xeonneon:
Closes #206 

## Introduced Changes
- Slight refactor of `build_package` to better handle exceptions throughout the build process
  - Separated out some functions to help with testability
  - Added a `_builder_package` helper method that returns the build and build log back to the parent `build_package` method, so that the Build status is updated accordingly and the build log is saved with any error messages
- Forcibly remove images as a precaution during image build cleanup

## Testing
- Added a couple of unit tests
  - Unit tests heavily rely on mocking
- Build images as usual


xeonneon:
After thinking about this PR over the weekend, I'm considering doing a larger refactor of the build process methods. I'm not very satisfied with this PR, and I think there is definitely a better way to organize this code to improve both the process and test-ability.

scott-taubman:
I had some pending comments that I won't bother to submit now if you're going to take another stab at this, but the one that probably remains relevant is that I wasn't crazy of some of what looked like Go style exception handling that you were doing, where you were returning error messages and then keying off of that rather than just handling actual exceptions.  The summary of what was going to be my comment is:

I'm a little torn on whether all of the exception handling in general is worth it or not. At the end of the day, pretty much any exception being raised is going to result in us marking the build status as ERROR and then recording the log, so it seems less error prone to just let the exceptions bubble back up to the build_package and let it handle recording the error. And then in all of these helper functions only really handle the exceptions that we can might be able to recover from or want to do something other than mark the build as failed.

functionary/builder/utils.py
```diff
@@ -105,6 +255,109 @@ def initiate_build(
...
+        build.status = Build.COMPLETE
+        _create_build_log(build, log)
+        _update_build_status(build)
+        logger.info(f"Build {build_id} COMPLETE")
+    except BuilderException as err:
+        logger.error(f"Encountered build error: {err}")
+        build.status = Build.ERROR
+        _create_build_log(build, err)
+        _update_build_status(build)
+    except DockerSocketConnectionError as err:
```
scott-taubman, line 306: Is there a reason why you wouldn't just record this as a build error like with the other exceptions?  It would still be useful to see that the build failed and why.

Also, if you do that, I think you can restructure this to just have a finally at the end that updates the build:

```python
    try:
        build, log = _build_package(build, build_resource)
        build.status = Build.COMPLETE
        logger.info(f"Build {build_id} COMPLETE")
    except Exception as err:
        log = str(err)
        build.status = Build.ERROR
        logger.error(f"Build {build_id} failed: {err}")
    finally:
        _create_build_log(build, log)
        _update_build_status(build)
```

I also wonder if having `_build_package` return the log is the correct thing to do, versus passing it an object that it can build up as the log.  Even just a list that it can keep adding entries to.  The reason I'm thinking that is because as is, if we don't make it to the end of `_build_package`, any of the potentially useful log entries up to that point are lost because we didn't reach the return.  If we instead were appending entries to a list inside `_build_package`, that list would be intact even if we never reach the return.  Then you could append the exception output to the rest of the log and we'd have the full picture of how far the build got before failing.

functionary/builder/utils.py
```diff
@@ -105,6 +255,109 @@ def initiate_build(
...
+        build_log: The log string of the entire build process
+
+    Raises:
+        DockerSockerConnectionError: Raised when connection to the docker socket
+        could not be established.
+    """
+    docker_client = _get_docker_client()
+
+    build.status = Build.IN_PROGRESS
+    _update_build_status(build)
```
scott-taubman, line 333: It seems like in all cases you set `build.status` and then immediately call `_update_build_status`.  Why not just do `build.in_progress()` here instead?  It's not clear to me what `_update_build_status` really buys us.

functionary/core/models/package.py
```diff
@@ -46,15 +47,22 @@ class Meta:
     def __str__(self):
         return self.name
 
+    def update_image_name(self, image_name: str) -> None:
```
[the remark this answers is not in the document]
xeonneon: Just a leftover from one of my previous iterations

functionary/builder/utils.py
xeonneon: Yea, that makes sense. That's originally what I was trying to accomplish with my Go-style iteration

functionary/builder/utils.py
xeonneon: Just leftover's from a previous iteration

functionary/builder/utils.py
```diff
@@ -87,7 +235,9 @@ def initiate_build(
                 environment=environment, name=package_definition.get("name")
             )
         except Package.DoesNotExist:
-            pass
+            package_obj = create_package_from_definition(
```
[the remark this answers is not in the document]
scott-taubman: I think deleting the package is likely to be problematic (as with deleting anything really).  Imagine this scenario:

A package exists and is active.  A bunch of tasking has happened against it.  It is then deactivated.  Someone attempts to publish a new version and the build fails.  You go to clean up the package because it's inactive.  All of the tasks against any of the functions in that package are deleted with it, which was probably not the intention.

If you want to add the active flag now, I would say just leave the package behind and inactive if the build fails.  We can figure out the best way to let people clean those up manually later.

functionary/builder/utils.py
```diff
@@ -105,6 +255,109 @@ def initiate_build(
...
+    """
+    docker_client = _get_docker_client()
+
+    build.status = Build.IN_PROGRESS
+    _update_build_status(build)
+
+    # Set variables necessary for build process
+    build_log = ""
+    workdir = _generate_path_for_build(build)
+    package: Package = build.package
```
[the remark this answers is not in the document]
xeonneon: Since I am guaranteeing that the package exists before the `build_image` method runs, I include the `Package` annotation

functionary/builder/utils.py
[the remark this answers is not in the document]
xeonneon: I ended up adding two distinct variables to prepare for that use case: `completed` and `enabled`

`completed` is meant to indicate that the Package has been built successfully at least once, and as such, can be rendered as a Package on the UI. Open to suggestions about changing this name.

`enabled` is meant for enabling or disabling a `Package` and it's related Functions. This would be what gets updated manually by the user when they wish to enable/disable a `Package`

The reason for this distinction is to separate out the state of "this package should be visible to clients" and "this package is available for use by clients"

functionary/builder/models/build.py
```diff
@@ -56,6 +56,26 @@ class Meta:
...
+        """Saves status as `COMPLETE`"""
+        self._update_status(self.COMPLETE)
+
+    def in_progress(self):
+        """Saves status as `IN_PROGRESS`"""
+        self._update_status(self.IN_PROGRESS)
+
+    def _update_status(self, status: str) -> None:
+        self.status = status
+        self.save()
```
kvig, line 77: Isn't this going to also save any other unsaved modifications to the build object?

functionary/builder/utils.py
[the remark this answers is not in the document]
scott-taubman: Would it not be better to update the model to be accurate?  It sounds like with your changes the expectation is that `build.package` should never be `None`.  If you update the field on the model to reflect that, then you don't need the annotation here because the type of `build.package` is actually guaranteed at that point.

functionary/builder/models/build.py
xeonneon: Yes. I can modify it to only save the status field. The other update status methods are just calling `self.save()`, so maybe I'll go back and update them to just save the status field in another PR

scott-taubman: Oh no, definitely don't do that.  The confusion that will be caused by having an object that is only partially consistent with the state of the database seems way more likely to be harmful than helpful.

functionary/builder/utils.py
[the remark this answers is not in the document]
xeonneon: I consolidated the two fields into just a `status` field. Although, I'm still not sure how I feel about those two fields being represented by one field, since the `PENDING` and `COMPLETED` fields are really only set once in a package's life cycle, and they are used by two separate entities

functionary/builder/models/build.py
xeonneon: Updated to just perform the save on the `status` field

functionary/builder/utils.py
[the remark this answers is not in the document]
scott-taubman: What are the "two separate entities" you are referring to here?

Also, to your point about it being set once and then never again, that same point also means that if you have a flag for "completed" it's basically `True` almost always and just sits there at `True` and will never again be `False` under any circumstance.  So it kind of just feels like a waste of space.  With the `status` field, once you move past the `COMPLETED` state you are no longer wasting space storing the fact that this thing successfully built once upon a time.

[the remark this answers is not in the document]
scott-taubman: To be clear, the storage space on the flag is not the concern.  It's just the fact that with the flag, once you set it to True, it kind of loses all value.  It also creates a theoretical situation where you could have something marked as not complete, but active, which obviously doesn't make sense and then you have to put in logic to check for etc.  Seems like a bad time all around.

functionary/builder/models/build.py
```diff
@@ -56,6 +55,29 @@ class Meta:
...
+
+    def in_progress(self):
+        """Saves status as `IN_PROGRESS`"""
+        self._update_status(self.IN_PROGRESS)
+
+    def _update_status(self, status: str) -> None:
+        """Update status of the build"""
+        if status != self.status:
+            self.status = status
+            self.save(update_fields=["status"])
```
scott-taubman, line 79: I commented elsewhere  on this, but please don't do this.  I have a hard time thinking of a situation where the benefit of doing this outweighs the pain it will likely cause someone one day.

functionary/builder/utils.py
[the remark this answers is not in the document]
xeonneon: The two entities are the Builder and the User. In my mind, the Builder needed a separate flag that it can manipulate to indicate that the package has been built, while the User should have their own flag for toggling whether the `Package` is enabled or disabled. Your point about the extra flag for the builder being a bit wasteful, I hadn't considered that, so I agree the single `status` field makes sense

functionary/builder/utils.py
```diff
@@ -105,6 +255,100 @@ def initiate_build(
...
+        build.error()
+    finally:
+        log = "\n".join(log_msgs)
+        _create_build_log(build, log)
+
+
+def _build_package(build: Build, build_resource: BuildResource, log: list[str]) -> None:
+    """Generate and build necessary resources to complete image build
+
+    Args:
```
scott-taubman, line 311: The docstring doesn't accurately reflect the function bahavior after recent changes.  Since this is an internal helper function, you can just have the one line summary docstring if you want.  Otherwise, let's update this to reflect the current state of things.

PR #249, opened 2023-03-03
248 - Add inidicator for how the tables are sorted

### LINKED ISSUE #248: Add a visual indicator for the column sorting
There is no indicator of what column is sorted on tables and in which direction.  That should be fixed.

scott-taubman:
Closes #248 

This PR adds some css styling to the already present "asc" and "desc" classes so that an indicator gets displayed on the column headers to show the current sorting.

## Test Instructions
* Go to any of the list views and sort.  You should see an indicator showing which way the sort is going.  Tasking and Builds are sorted by their date columns by default, so those show indicators even on the initial load.

### DIFF
--- a/functionary/ui/static/css/functionary.css
+++ b/functionary/ui/static/css/functionary.css
@@ -53,3 +53,22 @@ tr.htmx-swapping td {
 body div aside {
     max-width: fit-content;
 }
+
+/* table sorting indicators */
+th.asc::after {
+    content: '\f077';    /* chevron-up */
+    font-family: var(--fa-style-family,"Font Awesome 6 Free");
+    color: var(--bs-primary);
+    font-size: 0.75em;
+    line-height: .05em;
+    vertical-align: -.1em;
+    padding-left: .5em;
+}
+th.desc::after {
+    content: '\f078';    /* chevron-down */
+    font-family: var(--fa-style-family,"Font Awesome 6 Free");
+    color: var(--bs-primary);
+    font-size: 0.75em;
+    line-height: .05em;
+    padding-left: .5em;
+}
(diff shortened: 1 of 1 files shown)


functionary/ui/static/css/functionary.css
```diff
@@ -52,4 +52,23 @@ tr.htmx-swapping td {
...
+
+/* table sorting indicators */
+th.asc::after {
+    content: '\f077';    /* chevron-up */
+    font-family: var(--fa-style-family,"Font Awesome 6 Free");
+    color: var(--bs-primary);
+    font-size: 0.75em;
+    line-height: .05em;
+    vertical-align: -.1em;
+    padding-left: .5em;
```
xeonneon, line 65: My only suggestion would be to just consolidate the similar styles into one class, like:
```css
th.asc::after,
th.desc::after {
    font-family: var(--fa-style-family,"Font Awesome 6 Free");
    color: var(--bs-primary);
    font-size: 0.75em;
    line-height: .05em;
    vertical-align: -.1em;
    padding-left: .5em;
}
th.asc::after {
    content: '\f077';    /* chevron-up */
}
th.desc::after {
    content: '\f078';    /* chevron-down */
}
```

Also, for the `vertical-align`, is there any reason to not just let the element be vertically aligned with the middle of the element, rather than `-.1em`?

scott-taubman: Yeah, good call on consolidating these.

The `vertical-align` was an artifact from a some alignment wonkiness when I was using a different icon.  I removed it from the `desc` but forgot to take it out of the `asc` apparently.  It's gone now.

xeonneon (approved):

PR #250, opened 2023-03-06
179 Cleanup package tarball

lights7412:
This PR adds the option of -k or --keep to the publish command which will result in the demo tar file being stored in ~/.functionary/builds/. If the publish command is ran without the option (-k/--keep) it will simply be cleaned up after the operation is finished.

### Testing Instructions
` functionary package publish examples/demo` - Should result the tar file being cleaned up
`functionary package publish examples/demo --keep` - Should result in a tar file being stored in the ./functionary/builds/ folder
In addition 2 tests added to test_package.py

### DIFF
--- a/cli/.vscode/launch.json
+++ b/cli/.vscode/launch.json
@@ -1,0 +1,23 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "cli - attach",
+      "type": "python",
+      "request": "attach",
+      "connect": {
+        "host": "localhost",
+        "port": 56789
+      },
+      "pathMappings": [
+        {
+          "localRoot": "${workspaceFolder}",
+          "remoteRoot": "."
+        }
+      ]
+    }
+  ]
+}
--- a/cli/functionary/package.py
+++ b/cli/functionary/package.py
@@ -96,17 +96,16 @@ def create_cmd(ctx, language, name, output_directory):
 @package_cmd.command()
 @click.argument("path", type=click.Path(exists=True))
 @click.option(
-    "--keep", "-k", is_flag=True, help="Keep the tarball rather than cleaning it up"
+    "--keep",
+    "-k",
+    is_flag=True,
+    help="Keep build artifacts after publishing, rather than cleaning them up",
 )
 @click.pass_context
 def publish(ctx, path, keep):
     """
-    Create an archive from the project and publish to the build server.
+    Publish a package to make it available in the currently active environment.
 
-    This will create an archive of the files at the given path and
-    then publish them to the build server for image creation.
-    Use the -t option to specify a token or set the FUNCTIONARY_TOKEN
-    environment variable after logging in to Functionary.
     Use the -k option to keep the build artifacts
     (found in $HOME/.functionary/builds) after publishing,
     rather than cleaning it up.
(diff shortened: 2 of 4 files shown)


cli/.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Cli Debug",
```
scott-taubman, line 8: Let's call this `cli - attach`, both so that it matches the format of the other entries in the list and indicates that it's attaching to a thing rather than actually running a thing.

cli/functionary/package.py
```diff
@@ -95,30 +95,45 @@ def create_cmd(ctx, language, name, output_directory):
...
+
+    with tarfile.open(str(tar_path), "w:gz") as tar:
         tar.add(str(full_path), arcname="")
 
-    with open(tarfile_name, "rb") as upload_file:
-        click.echo(f"Publishing {str(tarfile_name)} package to {host}")
+    with open(tar_path, "rb") as upload_file:
+        click.echo(f"Publishing {str(tar_path)} package to {host}")
         response = post("publish", files={"package_contents": upload_file})
+        if keep:
```
scott-taubman, line 123: There's no need for a `pass` branch here.  You can just do:

```python
if not keep:
    os.remove(tar_path)
```

cli/tests/test_package.py
```diff
@@ -0,0 +1,47 @@
...
+
+def response_200(*args, **kwargs):
+    response = requests.Response()
+    response.status_code = 200
+    response._content = b'{"token": "somegreattoken"}'
+
+    return response
+
+
+# Calling publish with --keep ::: Resulting in tarball in the desired location
```
scott-taubman, line 20: These comments should be docstrings for the test function themselves.  So you'd remove this and then do:

```python
@pytest.mark.usefixtures("config")
def test_publish_with_keep(fakefs, monkeypatch):
    """Calling publish with --keep results in tarball not being removed"""
```

That kind of thing.

scott-taubman (changes requested):

cli/functionary/package.py
```diff
@@ -95,30 +95,45 @@ def create_cmd(ctx, language, name, output_directory):
 
 @package_cmd.command()
 @click.argument("path", type=click.Path(exists=True))
+@click.option(
+    "--keep", "-k", is_flag=True, help="Keep the tarball rather than cleaning it up"
```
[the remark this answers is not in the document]
lights7412: fixed

cli/functionary/package.py
```diff
@@ -95,30 +95,45 @@ def create_cmd(ctx, language, name, output_directory):
...
+        else:
+            os.remove(tar_path)
         id = response["id"]
         click.echo(f"Package upload complete\nBuild id: {id}")
 
 
+def get_tar_path():
+    tar_path = pathlib.Path.joinpath(pathlib.Path.home(), ".functionary")
+    pathlib.Path(tar_path, "builds").mkdir(parents=True, exist_ok=True)
+    return pathlib.Path.joinpath(tar_path, "builds", "demo.tar.gz")
```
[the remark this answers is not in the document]
lights7412: fixed

cli/tests/test_package.py
lights7412: fixed

cli/functionary/package.py
lights7412: fixed

cli/tests/test_package.py
```diff
@@ -0,0 +1,47 @@
...
+from click.testing import CliRunner
+
+from functionary.config import save_config_value
+from functionary.package import get_tar_path, publish
+
+
+def response_200(*args, **kwargs):
+    response = requests.Response()
+    response.status_code = 200
+    response._content = b'{"token": "somegreattoken"}'
```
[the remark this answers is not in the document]
lights7412: fixed

cli/.vscode/launch.json
lights7412: fixed

PR #252, opened 2023-03-06
#201 - workflow detail

obr42:

Converts Workflow detail page into bootstrap. All functionality on the page should work and by styled, including modals. Step table rows can be sorted with up/down arrows. Made a note on the UI ToDo to look into drag/drop as alternate for table row arrows.

Added missing breadcrumbs to Workflow Detail page. A little code cleanup / remove necessary classes.

Main title icon seemed excessively large - decreased size. Changed a few buttons to `primary` from `success` color.

### DIFF
--- a/functionary/ui/templates/partials/workflows/step_list.html
+++ b/functionary/ui/templates/partials/workflows/step_list.html
@@ -24,7 +24,7 @@
                                             {% if previous is None %}hidden{% endif %}
                                             hx-post="{% url 'ui:workflowstep-move' workflow.pk step.pk %}"
                                             hx-target="#workflow-steps"
-                                            hx-swap="OuterHTML"
+                                            hx-swap="outerHTML"
                                             hx-vals='{"next": "{{ previous.id }}"}'/>
                                     <button class="btn btn-small text-primary fa fa-chevron-down border-0"
                                             id="step-{{ step.id }}-move-down"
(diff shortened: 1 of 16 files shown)


functionary/ui/templates/partials/workflows/step_list.html
```diff
@@ -1,84 +1,77 @@
...
+                            {% with previous=step.previous next=step.next %}
+                                <div class="d-flex {% if previous is None %}justify-content-end{% elif next is None %}justify-content-start{% else %}justify-content-between{% endif %}">
+                                    <button class="btn btn-small text-primary fa fa-chevron-up border-0"
+                                            id="step-{{ step.id }}-move-up"
+                                            type="button"
+                                            title="Move Step Up"
+                                            {% if previous is None %}hidden{% endif %}
+                                            hx-post="{% url 'ui:workflowstep-move' workflow.pk step.pk %}"
+                                            hx-target="#workflow-steps"
+                                            hx-swap="OuterHTML"
```
scott-taubman, line 27: There's a typo / bug here that was there before, but just sort of happened to be masked.  This is case sensitive and needs to be `outerHTML`.  As is, the value isn't valid and reverts to the default of `innerHTML`, which results in the step list shifting a little bit when you first move a step.

scott-taubman (approved):

PR #254, opened 2023-03-08
#201 - schedule detail

obr42:
Does not close ticket - more work to do.

Converts Schedule detail page to Bootstrap. Also converts edit and create schedule pages (since they are the same HTML file).  Added missing breadcrumb to create/edit page. Status on detail page changes color depending on status.

Change `error` and `errorlist` django form css classes to use boostrap's `text-danger` color

Minor code cleanup. Condense gaps between sections on all pages. Reduce number of rows in textarea (is able to expand with corner drag).

UPGRADES TO BOOTSTRAP 5.3 this will allow some additional color CSS variables (used in this PR) and a dark theme (future work)


functionary/ui/templates/core/scheduledtask_detail.html
```diff
@@ -1,77 +1,63 @@
...
-                        <span class="fa fa-pencil-alt"></span>
-                    </button>
-                </a>
-            </p>
+    <div class="hstack gap-3 pt-2 pb-4">
+        <h1>
+            <i class="fa fa-clock"></i>
+            {{ scheduledtask.name }}
         </h1>
+        <i class="fa fa-pencil-alt fa-lg text-info singletonActive"
```
scott-taubman, line 17: What is `singletonActive`? Is that a holdover from bulma or is it doing something here?

functionary/ui/templates/forms/scheduled_task/scheduled_task_edit.html
```diff
@@ -1,233 +1,119 @@
...
-                                    <div class="errorlist">{{ form.function.errors }}</div>
-                                </div>
-                                <div class="column is-one-third">
-                                    {% if update %}
-                                        <div class="control">{% render_field form.function disabled=True %}</div>
-                                    {% else %}
-                                        <div class="control">{{ form.function }}</div>
+                        <div>
+                            <strong class="fs-4"><i class="fa fa-calendar fa-sm fa-fw pe-1"></i>Crontab Schedule</strong>
+                            <div class="ps-4 py-2 vstack gap-4">
```
scott-taubman, line 69: We probably want to do something other than `vstack` here.  The input fields here are going to take a max of like 4 characters, so having them take the entire width of the form feels excessive.  Just switching to `hstack` seems okay, but has two problems of its own:

* The field labels get cutoff
* The error message placement no longer really works

Those both feel like solvable problems though and it would clean up the page a lot I think.

functionary/ui/templates/forms/scheduled_task/scheduled_task_edit.html
```diff
@@ -1,233 +1,119 @@
...
+                                {% for formItem in form %}
+                                    {% if 'scheduled' in formItem.id_for_label %}
```
scott-taubman, line 71: Minor thing, but for consistency sake let's stick to snake_case rather than camelCase for variable names.

functionary/ui/templates/core/scheduledtask_detail.html
scott-taubman: Okay, that probably makes sense then.  Carry on.

scott-taubman (approved):

PR #257, opened 2023-03-09
#222 - Added error handling to package yaml extraction

### LINKED ISSUE #222: Builder: Publish endpoint can throw a 500 when package.yaml is invalid
If the package.yaml is malformed in some specific ways the API will raise an exception and throw a 500 rather than handling the error properly and returning a 400.  For example, if someone were attempting to add a new function, but doesn't properly indent the start of the entry:

```yaml
    - name: num_chars
      summary: Returns the number of characters in given file
      display_name: Number of Characters
      description: Returns the number of characters in given file
      return_type: integer
      parameters:
        - name: file
          display_name: File
          summary: File to measure
          description: Input file that will be measured
          type:  file
          required: true
      # This is under-indented
      - name: parameter_types
        summary: Test for parameter type form rendering and passthrough values
        display_name: Parameter Types
        parameters:
          - name: boolean
            type: boolean
            required: False
```

All such errors should be handled gracefully with as useful of a response back to the user as we can muster.

xeonneon:
Closes #222 

## Introduced Changes
- Added more `try`/`excepts` to parts of the `extract_package_definition` method in the builder
  - Handle the YAML parse issues to prevent `500` error. Now returns a `400`

## Testing
Create a new package with a malformed package.yaml:
```yaml
version: 1.0
package:
  name: "invalid"
  display_name: Functionality Demo
  summary: Provides functions that demo / test various available features
  language: "python"
  functions:
    - name: output_json
      summary: Demonstrates JSON output rendering
      display_name: Output JSON
      description:
        'Takes in a JSON blob and then simply returns that to demonstrate how
        the JSON rendering works in the UI. To see table rendering, the input
        should be a list formatted something like: [ { "param1": "value1",
        "param2": "value2" }, { "param1": "value3", "param2": "value4" } ]'
      parameters:
        - name: input
          summary: The JSON blob to render
          type: json
          required: true
      return_type: json
  - name: output_text
    summary: Demonstrates text output rendering
    display_name: Output Text
    description:
      Takes in a string and then simply returns that to demonstrate how string
      results rendering works in the UI. CSV formatted text will provide the
      option to be rendered as a table.
    parameters:
      - name: input
        summary: The text to render
        type: text
        required: true
    return_type: string
```

## Notes
In your `package.yaml`, if `.package.functions` is just an empty dictionary, no exception is raised. We probably should define a schema for the packages once we finalize all of the fields, such that incorrect field types are handles. This will also help combat some of the auto type casting YAML will perform, such as turning `yes` into a boolean in fields that are meant to be strings.

### DIFF
--- a/functionary/builder/tests/test_utils.py
+++ b/functionary/builder/tests/test_utils.py
@@ -1,4 +1,5 @@
 import pytest
+from yaml import YAMLError
 
 from builder import utils
 from core.models import Function, Package, Team, User
@@ -13,6 +14,11 @@ def user():
 @pytest.fixture
 def team():
     return Team.objects.create(name="team")
+
+
+@pytest.fixture
+def package_contents():
+    return bytes("example package contents", encoding="utf-8")
 
 
 @pytest.fixture
@@ -119,3 +125,37 @@ def test_delete_removed_function_parameters(function1):
 
     assert function1.parameters.count() == 1
     assert function1.parameters.filter(name="param1").exists()
+
+
+def test_invalid_package_yaml(mocker, package_contents):
+    class TarFile:
+        def close():
+            return
+
+    def mock_tarfile(_bytes):
+        return TarFile
+
+    def mock_missing_package_yaml(_package_yaml):
+        raise KeyError("missing package yaml")
+
+    def mock_invalid_package_yaml(_package_yaml):
+        raise YAMLError
+
+    def mock_invalid_package(_package_yaml):
+        raise utils.InvalidPackage
+
+    mocker.patch("builder.utils._get_tarfile", mock_tarfile)
+    mocker.patch("builder.utils._extract_package_definition", mock_missing_package_yaml)
+
+    with pytest.raises(utils.InvalidPackage):
+        utils.extract_package_definition(package_contents)
+
+    mocker.patch("builder.utils._extract_package_definition", mock_invalid_package_yaml)
+
+    with pytest.raises(utils.InvalidPackage):
+        utils.extract_package_definition(package_contents)
+
+    mocker.patch("builder.utils._extract_package_definition", mock_invalid_package)
+
+    with pytest.raises(utils.InvalidPackage):
+        utils.extract_package_definition(package_contents)
--- a/functionary/builder/utils.py
+++ b/functionary/builder/utils.py
@@ -3,7 +3,8 @@ import json
 import logging
 import os
 import shutil
-import tarfile
+from tarfile import ReadError, TarFile
+from tarfile import open as open_tarfile
 from typing import List
 from uuid import UUID
 
@@ -14,7 +15,6 @@ from django.db import transaction
 from django.template.loader import get_template
 from docker.errors import APIError, BuildError, DockerException
 from yaml import YAMLError, safe_load
-from yaml.parser import ParserError
 
 from core.models import Environment, Function, FunctionParameter, Package, User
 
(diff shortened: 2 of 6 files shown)


functionary/builder/utils.py
```diff
@@ -33,34 +34,49 @@ def extract_package_definition(package_contents: bytes) -> dict:
...
+        tarball: The opened tarfile in read mode
 
-    package_definition = yaml.safe_load(package_definition_io.read())
-    close_files()
+    Returns:
+        package_contents: A dictionary representation of the package yaml
+
+    Raises:
+        InvalidPackage: Raised when the package.yaml is not a regular file
+        ParseError: Raised when there is an error parsing the package yaml
```
scott-taubman, line 72: I'm not clear on the distinction between `ParseError`  (which I assume is actually `ParserError`) and `YAMLError`.  I think the former is just a variation of the latter and you can just handle `YAMLError` and be fine.  Based on the documentation it seems like the intention is that `YAMLError` is the catch-all public facing exception that users are expected to handle.

scott-taubman (changes requested):
The changes mostly look good.  I made one minor comment about the yaml exceptions, but the other thing I'll suggest is that we should add at least one unit test here.  There's an easily reproducible problem that these changes are intending to fix.  We should write a test that induces the problem and see that these changes do in fact fix it.

functionary/builder/utils.py
xeonneon: I originally had separate except blocks before I consolidated them into one. The `ParseError` was just a leftover from that implementation

functionary/builder/tests/test_utils.py
```diff
@@ -119,3 +125,37 @@ def test_delete_removed_function_parameters(function1):
 
     assert function1.parameters.count() == 1
     assert function1.parameters.filter(name="param1").exists()
+
+
+def test_invalid_package_yaml(mocker, package_contents):
```
scott-taubman, line 130: These tests are a little bit in the weeds in that they feel like they're testing specific implementation details.  I pushed up a test in `views/test_publish.py` that is more of what I had in mind when I suggested adding tests.  That is, it reproduces the actual bug from the API user perspective: malformed yaml is provided during the publish and the server throws an error rather responding with a 400.  That test blows up if run against the code prior to your changes.  With your changes, the test passes.  When addressing bugs, I think that's generally the style of testing that's most useful because it specifically confirms that the the bug has been addressed, as opposed to confirming that some internal implementation detail has improved in a way that only sort of implies the bug has been fixed.

functionary/builder/utils.py
```diff
@@ -33,34 +34,54 @@ def extract_package_definition(package_contents: bytes) -> dict:
...
-
-    try:
-        package_definition_io = tarball.extractfile("package.yaml")
     except KeyError:
-        close_files()
         raise InvalidPackage("package.yaml not found")
+    except (InvalidPackage, YAMLError):
+        raise InvalidPackage("package.yaml is invalid YAML")
+    finally:
+        tarfile.close()
```
scott-taubman, line 57: Something I only noticed when adding the view test is that this line will blow up if the `tarfile = _get_tarfile(package_contents_io)` is what throws the exception, since `tarfile` is not set in that case.

xeonneon: Fixed

functionary/builder/tests/test_utils.py
xeonneon: Gotcha. I'll start including integration and regression tests with my unit tests

scott-taubman (approved):

PR #260, opened 2023-03-13
Added ability to refresh the login token.

### LINKED ISSUE #251: User Token Management
Currently API tokens are not manageable or easily retrievable for users.  Especially in the case of users that are using 3rd party auth, as they cannot just authenticate with the API and hit the `/api-token-auth` endpoint.

We should make the tokens retrievable via the UI in a menu that's similar to the "account connections" page.  It may even be appropriate to just update the "account connections" page to be a more general user settings page with different sections, one of which is the token management.

Users should be able to retrieve their current token as well as generate a new token to replace it (invalidating the previous token in the process).  This is far from the most robust token management implementation as it's still just a single token that has the user's full access associated with it, but the goal here is to provide a means of retrieving and changing the token and not a broader overhaul of the underlying token system.

kvig:
This PR adds the ability to see and refresh the users token. This is useful if the user needs to remove a compromised token, or if you need to get a login token to make API calls over HTTP.

This change involved moving the social account connections page to a partial and adding a new account details page that also shows the token.  The social account delete view from django-allauth had to be circumvented because it would always redirect to a fixed URL instead of listening to the `next` parameter.

Closes #251

### DIFF
--- a/functionary/ui/templates/partials/token.html
+++ b/functionary/ui/templates/partials/token.html
@@ -1,18 +1,43 @@
 <div id="token">
     <strong class="fs-4"><i class="fa fa-id-badge fa-sm fa-fw pe-1"></i>Token</strong>
-    <div class="col-9">
-        <form >
-            {% csrf_token %}
-            <span class="col-7" data-token="{{ token.key }}">{{ token.key }}</span>
-            {% comment %} <span class="col-2"/> {% endcomment %}
-            <button class="btn btn-sm border-0 btn-outline"
-                    type="submit"
-                    title="Regenerate Token"
-                    hx-target="#token"
-                    hx-swap="outerHTML"
-                    hx-post="{% url 'ui:token-refresh' %}">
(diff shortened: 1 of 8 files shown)


functionary/ui/templates/partials/token.html
```diff
@@ -0,0 +1,18 @@
+<div id="token">
+    <strong class="fs-4"><i class="fa fa-id-badge fa-sm fa-fw pe-1"></i>Token</strong>
+    <div class="col-9">
+        <form >
+            {% csrf_token %}
+            <span class="col-7" data-token="{{ token.key }}">{{ token.key }}</span>
```
scott-taubman, line 6: It might be good to just show •••••••••••••• or something similar and then have a button for revealing the token, rather than just showing it by default.  It is a secret after all.

kvig: Added a button to swap between masked and unmasked.

PR #263, opened 2023-03-15
#201 - team detail

obr42:
Not quite close ticket, there are still some misc files that need to be converted.

Converts team detail page to bootstrap. Relies on some of the environment detail work. Removes a few outdated perm check variables for environment/teams and updates to the new `request.session` format - corresponding python definitions removed.

A little bit of general cleanup.

### DIFF
--- a/functionary/ui/templates/partials/teams/team_management.html
+++ b/functionary/ui/templates/partials/teams/team_management.html
@@ -26,7 +26,7 @@
                                             hx-target="#user-modal">
                                     </button>
                                 {% endif %}
-                                {% if request.session.user_can_delete_team %}
+                                {% if request.session.user_can_delete_userrole %}
                                     <button class="btn btn-small text-danger fa fa-trash"
                                             title="Delete"
                                             type="button"


functionary/ui/templates/core/team_detail.html
```diff
@@ -1,60 +1,48 @@
...
+                <div class="ms-4 my-2">
+                    <div id="users" class="list-group list-group-flush">
                         {% for u in user_details %}
-                            <li class="block ml-4">
-                                <p>
-                                    {{ u.first_name }} {{ u.last_name }} <span class="has-text-grey-light">({{ u.username }})</span>
-                                </p>
-                            </li>
+                            <p class="list-group-item fw-bold mb-0">
+                                {{ u.first_name }} {{ u.last_name }}
```
scott-taubman, line 36: This was wrong before and not a result of these changes, but might as well fix it for now.  The `u` here is not the user object, it's some other object that contains role info in addition to the user.  So each `u.` would have to be `u.user.`.

I wonder if the better thing to do here would be remove the permissions check here and then just wrap the various buttons in the `team_management.html` with the permissions checks.  That way we don't have separate rendering logic for this info, it's just that the editing is disabled if you don't have access.

functionary/ui/templates/partials/variable_row.html
```diff
@@ -10,15 +10,15 @@
         {% endif %}
     </td>
     <td>
-        {% if var_update %}
+        {% if request.session.user_can_update_variable %}
```
scott-taubman, line 13: Seeing this made me realize that the "Add" button in `variables_section.html` is not wrapped in this same check, so while the edit buttons for existing entries are not present, that button for creating new entries is.

functionary/ui/templates/partials/environments/environment_management.html
```diff
@@ -20,44 +20,45 @@
...
-                                    </button>
-                                {% endif %}
-                                {% if u.environment_user_role_id %}
-                                    <button class="btn btn-small text-danger fa fa-trash"
-                                            title="Delete"
-                                            type="button"
-                                            hx-post="{% url 'ui:environmentuserrole-delete' environment_id u.environment_user_role_id %}"
-                                            hx-confirm="Are you sure you want to remove {{ u.user.username }}'s explicit role from this environment?">
-                                    </button>
+                                {% if request.session.user_can_update_environment %}
```
scott-taubman, line 23:
```suggestion
                                {% if request.session.user_can_update_userrole %}
```

functionary/ui/templates/partials/environments/environment_management.html
```diff
@@ -20,44 +20,45 @@
...
-        <div id="update-user-modal"></div>
-        <button class="btn btn-primary fa-plus"
-                type="button"
-                hx-target="#add-user-modal"
-                hx-trigger="click"
-                hx-get="{% url 'ui:environmentuserrole-create' environment_id %}">
-            Add
-        </button>
+        <div id="user-modal"></div>
+        {% if request.session.user_can_create_environment %}
```
scott-taubman, line 54:
```suggestion
        {% if request.session.user_can_create_userrole %}
```

functionary/ui/templates/partials/teams/team_management.html
```diff
@@ -1,48 +1,56 @@
...
+                    </tr>
+                </thead>
+                <tbody>
+                    {% for u in user_details %}
+                        <tr>
+                            <td>{{ u.user.first_name }} {{ u.user.last_name }}</td>
+                            <td>{{ u.user.username }}</td>
+                            <td>{{ u.role }}</td>
+                            <td>
+                                {% if request.session.user_can_update_team %}
```
scott-taubman, line 21:
```suggestion
                                {% if request.session.user_can_update_userrole %}
```

functionary/ui/templates/partials/teams/team_management.html
```diff
@@ -1,48 +1,56 @@
...
+                                    <button class="btn btn-small text-info fa fa-pencil-alt"
+                                            title="Edit user"
+                                            type="button"
+                                            hx-get="{% url 'ui:teamuserrole-update' team_id u.team_user_role_id %}"
+                                            hx-target="#user-modal">
+                                    </button>
+                                {% endif %}
+                                {% if request.session.user_can_delete_team %}
```
scott-taubman, line 29:
```suggestion
                                {% if request.session.user_can_delete_userrole %}
```

functionary/ui/templates/partials/teams/team_management.html
```diff
@@ -1,48 +1,56 @@
...
+                                    </button>
+                                {% endif %}
+                            </td>
+                        </tr>
+                    {% endfor %}
+                </tbody>
+            </table>
+        </div>
+        <div id="user-modal"></div>
+        {% if request.session.user_can_create_team %}
```
scott-taubman, line 46:
```suggestion
        {% if request.session.user_can_create_userrole %}
```

scott-taubman (approved):

PR #273, opened 2023-03-20
# 201 - final bootstrap implementation

### LINKED ISSUE #201: Switch to a better supported CSS component library

obr42:
Closes #201 

Catch-all for some remaining files that needed to be converted. Removed some OBE files. Update data sections to use heading elements - should be minimal net visual change, but makes pages better accessible to screen readers. Update `Schema` section to look more like other sections. Also knocks out some points from the UI ToDo list: 

- Tooltip/title at least when in icon only size so users know what they are clicking
- Highlight active page
- Consistent buttons (title, type, icon vs words)

Could not remove the main bulma file from static as the /admin page is still using it.


functionary/ui/templates/registration/login.html
```diff
@@ -11,19 +11,17 @@
...
+    {% csrf_token %}
+    <div class="mb-3">
+      {{ form.username.label_tag }}
+      {{ form.username }}
+    </div>
+    <div class="mb-3">
+      {{ form.password.label_tag }}
+      {{ form.password }}
+    </div>
+f    <button class="btn btn-primary" type="submit" title="Login">Login</button>
```
scott-taubman, line 24: Looks like a stray "f" made its way into the start of this line.

functionary/ui/scss/custom.scss
```diff
@@ -32,9 +35,10 @@ $dark: #343E40;
   ------------------*/
 $enable-rfs: false; // Disable Responsive Font Sizes
 
-$navbar-dark-color: $secondary;
-$navbar-dark-hover-color: #D3DAE4; // Slight deviation from default for compliance
+$navbar-dark-color: rgba($gray-200, .75);
+$navbar-dark-hover-color: rgba($white, .75);
```
scott-taubman, line 39: This hover color seems pretty indistinguishable from the `navbar-dark-color`.  Two options that look okay to me just from messing around:

* Go darker, with .35 opacity.
* Go full $white.  Adding "fw-semibold" in addition to "active" could help slightly differentiate the currently active from the hover over if we go this route.

PR #280, opened 2023-03-23
#274 - Rework management pages

### LINKED ISSUE #274: Rework accessing management pages
The environment and team list views shouldn't exist. Instead of having links to those list views in the nav, there should be links somewhere for "manage environment" and "manage team" that bring you directly to the detail pages for the active environment / team of the active environment.

We've talked about it maybe being a "..." next to the environment dropdown, but that may or may not work in reality.  If not, we'll need some other natural place to put that navigation.

obr42:
Closes #274 

Remove Environment List and Team List table pages and all references. Refactor the environment select to be a split-button that still allows users to change environment, but also navigate to the specific team and environment page.

Added `menu` role to all selects for compliance.

Had to add Bootstrap's JS file to enable drop-down capability. Used the `bundle.min.js` file that includes PopperJS (a dependency library).


functionary/ui/templates/forms/environment_selector.html
```diff
@@ -1,25 +1,55 @@
...
+                </button>
+                <div class="dropdown-menu shadow-lg dropdown-menu-dark">
+                    <a href="{% url 'ui:environment-detail' pk=environment_id %}"
+                       class="dropdown-item"
+                       title="Manage Environment">
+                        <i class="fa fa-lg fa-building fa-fw me-1"></i>Manage Environment</a>
+                    <a href="{% url 'ui:team-detail' pk=env.team.id %}"
+                       class="dropdown-item"
+                       title="Manage Team">
+                        <i class="fa fa-lg fa-users fa-fw me-1"></i>Manage Team</a>
```
scott-taubman, line 50: Minor nit, but the `fa-lg` feels too big here.  Playing around with it I kind of think going all the way down to `fa-sm` looks best.

PR #282, opened 2023-03-24
#265 - hide empty sections

### LINKED ISSUE #265: Hide sections with no data
Consistently hide a section if there is no content in that section. The if statement should be around the main div for the section. This is especially important for for-loops as they will throw an error if the list is undefined

obr42:
Closes #265 

Completely hides sections that are empty by wrapping optional/possibly undefined variable displays in an if statement.


functionary/ui/templates/core/environment_detail.html
```diff
@@ -21,8 +21,12 @@ <h3 class="fw-semibold">
                 <a href="{% url 'ui:team-detail' pk=environment.team.id %}">{{ environment.team.name }}</a>
             </p>
         </div>
-        {% include "partials/environments/environment_management.html" %}
-        {% include "partials/variables_section.html" with parent_object=environment parent_id=environment.id %}
+        {% if user_details %}
```
kvig, line 24: These both still need to show up so you can add users and variables to the environment

functionary/ui/templates/core/team_detail.html
```diff
@@ -13,22 +13,28 @@ <h1 class="mt-2 mb-4">
...
+                            </a>
+                        {% endfor %}
+                    </div>
                 </div>
             </div>
-        </div>
-        {% include "partials/teams/team_management.html" %}
-        {% include "partials/variables_section.html" with parent_object=team parent_id=team.id %}
+        {% endif %}
+        {% if user_details}
```
kvig, line 33: We need to always be able to add users and variables if there aren't any.

functionary/ui/templates/core/workflow_detail.html
```diff
@@ -21,23 +21,29 @@ <h2>
...
+        {% if workflow.parameter %}
+            <div>
+                <h3 class="fw-semibold">
+                    <i class="fa fa-list fa-sm fa-fw me-1"></i>Parameters
+                </h3>
+                {% include "partials/workflows/parameter_list.html" %}
+            </div>
+        {% endif %}
+        {% if workflow.ordered_steps %}
+            <div>
```
kvig, line 41: I would show "No steps" or "None" since a workflow is expected to have steps.  Hiding the description and parameters is fine since they're not necessary to exist.

kvig (changes requested):

kvig (approved):

PR #294, opened 2023-03-28
#288 - Refactor table filters

### LINKED ISSUE #288: Improve styling of filters, especially as window resizes

obr42:
Closes #288 

Refactor filters so they are a static sidebar at "wide" screen size and a button at "narrow" screen size. The button opens an off-canvas sidebar that contains the filter inputs.

Add `<hr>` between breadcrumbs and content. Adjust main content width to make tables wider without making input forms look too bad. All core pages should have the same amount of header space (content starts in same spot).

Fix bug where Schedule Task status was not on the correct line.

Note in UI TODO for adding indicator on "narrow" screen filter button that filters are applied and clear all filters button.


functionary/ui/templates/builder/build_detail.html
```diff
@@ -3,61 +3,66 @@
     <div id="build_detail"
          hx-swap-oob="true"
          {% if not completed %} hx-get="{% url 'ui:build-detail' build.id %}?poll=true" hx-trigger="every 15s"{% endif %}>
-        <nav aria-label="breadcrumb">
-            <ol class="breadcrumb">
+        <nav aria-label="breadcrumb mb-1">
```
scott-taubman, line 6: The margin between the breadcrumb and the `<hr>` is different on the detail page versus the list page, which results in a subtle shift as you navigate between pages.  Here's a snippet from the two jammed together with the text aligned.  You can see slightly bigger gap between the text and the `<hr>` on the left.

image

functionary/ui/templates/partials/list_filter.html
```diff
@@ -1,20 +1,49 @@
...
+                <form hx-get={{ filter.request.path }} hx-trigger="submit delay:500ms" hx-push-url="true" hx-headers='{"x-filter": "filter"}' hx-swap="outerHTML" hx-target="#table-list-block">
+                    {% for field in filter.form %}
+                        <div class="form-floating mb-3">
+                            {% with WIDGET_ERROR_CLASS="is-invalid" %}
+                                {% if field.widget_type == "select" %}
+                                    {% render_field field class="form-select form-select-sm" onchange="requestSubmit()" role="menu" %}
+                                {% elif field.widget_type == 'datetime' %}
+                                    {% render_field field class="form-control form-control-sm" onchange="requestSubmit()" %}
+                                {% else %}
+                                    {% render_field field class="form-control form-control-sm" onchange="requestSubmit()" %}
```
scott-taubman, line 33: These fields lost the autosubmit after the delay.  So you have to actually change the focus now for the filter to be applied.  Was that intentional?  If so, I feel like the old behavior was preferable.

functionary/ui/templates/partials/list_filter.html
```diff
@@ -1,20 +1,49 @@
...
+            </h2>
+            <button type="button"
+                    class="btn-close btn-close-white"
+                    data-bs-dismiss="offcanvas"
+                    data-bs-target="#filterOC"
+                    aria-label="Close">
+            </button>
+        </div>
+        <div class="offcanvas-body border rounded">
+            <div class="vstack p-2 form-fill">
```
scott-taubman, line 20: This will surely rank highly on my list of most nitpicky comments, but can we make this `px-2 pt-2` instead?  That trims a little of the hanging fat off the bottom of the filter box.

scott-taubman (commented):
A few mostly minor comments.

One thing that playing around with it made me realize we neglected is some way to clear the filters.  We can address that here or in a follow-on, but it seemed worth pointing out.

functionary/ui/templates/core/workflow_list.html
```diff
@@ -1,24 +1,39 @@
...
+                   type="button"
+                   title="Create New Workflow"
+                   hx-target="body"
+                   hx-swap="beforeend"
+                   hx-get="{% url 'ui:workflow-create' %}">
+                    Create Workflow
+                </a>
+            {% endif %}
+            {% if filter %}
+                <button class="btn btn-primary btn-sm d-xl-none py-1 h-75"
```
scott-taubman, line 21: If this button is going to be the same everywhere, it might be worth turning it into its own partial and including it.  That way if it needs to be tweaked it only has to happen in one place.

PR #300, opened 2023-03-29
#296 - Show "no parameters"

### LINKED ISSUE #296: Do not hide Parameters sections, instead say "No parameters"

obr42:
Closes #296 

Parameters should display `no parameters` instead of the whole section being hidden


functionary/ui/templates/core/scheduledtask_detail.html
```diff
@@ -59,16 +59,18 @@ <h3 class="fw-semibold">
...
+        <div>
+            <h3 class="fw-semibold">
+                <i class="fa fa-list fa-sm fa-fw me-1"></i>Parameters
+            </h3>
+            <div class="my-2">
+                {% if scheduledtask.parameters %}
                     <ol class="json-container wrap-json" id="json-result" />
-                </div>
+                {% else %}
+                    <p class="font-monospace">No parameters defined</p>
```
scott-taubman, line 70: The wording should be consistent between this and task_detail.html.  Probably just "No parameters" is fine.

Also, everything else has `ms-4`, so the text here feels misaligned.

scott-taubman (commented):
The schedule edit page (`scheduled_task_edit.html`) could use the "No parameters" text as well.  In the case where there is a form, you can additionally check `task_parameter_form.fields` to determine if the form should be rendered.  If not, render the "No parameters" instead.

functionary/ui/templates/forms/scheduled_task/scheduled_task_edit.html
```diff
@@ -110,7 +110,12 @@ <h3 class="fw-semibold">
                             <div class="font-monospace ms-4 my-2" id="function-parameters">
                                 {% comment %} Render function param form if available. Else, lazy load. {% endcomment %}
                                 {% if task_parameter_form %}
-                                    {{ task_parameter_form }}
+                                    {% if task_parameter_form.fields %}
+                                        has form
```
scott-taubman, line 114:
```suggestion
```

Looks like a line from some debugging or something snuck in here.

PR #305, opened 2023-03-30
Filter the UI selection of Functions and Packages to only the active objects.

### LINKED ISSUE #259: Appropriately handle disabled packages
Packages now have a status field and can be marked as disabled.  We need to properly handle disabled packages in various spots:

* Querysets need to be updated so that disabled packages are not shown in various places.
* Function behavior needs to be updated so that a Function is not considered active if it's package is disabled.
  * We do not want to actually set active=False on functions when a package is disabled because that would be a mess for if / when the package is re-enabled.

kvig:
The UI now filters out inactive Functions as well as active Functions of disabled Packages. When disabling a Package, the Scheduled Tasks for the Functions will also be paused. When enabling a Package, Scheduled Tasks for Functions will *not* be re-activated, this is a manual step.

Disabled Functions do not appear in the Function list when updating an existing Workflow Step or Scheduled Task. When the Package is re-enabled or the Function reactivated, it will show up in those lists again.

There were new tests added to cover these changes. You can manually test these changes out in the UI by starting a Django shell and manually deactivating/activating a Function or disabling/enabling a Package.
```python
from core.models import Package, package
ps = Package.active_objects.all()
ps
<QuerySet [<Package: calculator>, <Package: demo>]>
p=ps[1]
p.disable()
p.enable()
```

** NOTE **: All Functions will probably show up as inactive because the existing Packages are in the "COMPLETE" state. The Packages can be either republished or manually enabled in the shell as above.

Fixes #259.

### DIFF
--- a/functionary/core/models/function.py
+++ b/functionary/core/models/function.py
@@ -5,7 +5,7 @@ from django.contrib.contenttypes.fields import GenericRelation
 from django.core.exceptions import ValidationError
 from django.db import models
 
-from core.models.package import ENABLED
+from core.models.package import PACKAGE_STATUS
 from core.utils.parameter import get_schema
 
 
@@ -19,10 +19,17 @@ def list_of_strings(value):
 
 
 class ActiveFunctionManager(models.Manager):
-    """Manager for filtering out disabled packages."""
+    """Manager that filters out inactive Functions.
+
+    An inactive Function is a Function explicitly marked inactive
+    or an active Function of an inactive Package"""
 
     def get_queryset(self):
-        return super().get_queryset().filter(active=True, package__status=ENABLED)
+        return (
+            super()
+            .get_queryset()
+            .filter(active=True, package__status=PACKAGE_STATUS.ACTIVE)
+        )
 
 
 class Function(models.Model):
@@ -91,23 +98,12 @@ class Function(models.Model):
         """Deactivate the function and pause any associated scheduled tasks"""
         self.active = False
         self.save()
-        self.update_scheduled_tasks(enable=False)
+        self.pause_scheduled_tasks()
 
-    def update_scheduled_tasks(self, enable: bool) -> None:
-        """Activates all related scheduled tasks if enable is True,
-        otherwise pause the scheduled tasks.
-
-        Args:
-            enable: If true, activates the scheduled task, otherise pauses.
-
-        Returns:
-            None
-        """
+    def pause_scheduled_tasks(self) -> None:
+        """Pauses all scheduled tasks."""
         for scheduled_task in self.scheduled_tasks.all():
-            if enable:
-                scheduled_task.activate()
-            else:
-                scheduled_task.pause()
+            scheduled_task.pause()
 
     @property
     def parameters(self):
@@ -125,6 +121,7 @@ class Function(models.Model):
         """Function definition schema"""
         return get_schema(self)
 
+    @property
     def is_active(self) -> bool:
-        """Returns true if active and the package isn't disabled"""
-        return self.active and self.package.enabled
+        """Returns true if this Function and its Package are both active."""
+        return self.active and self.package.is_active
--- a/functionary/core/models/function.py
+++ b/functionary/core/models/function.py
@@ -5,6 +5,7 @@ from django.contrib.contenttypes.fields import GenericRelation
 from django.core.exceptions import ValidationError
 from django.db import models
 
+from core.models.package import ENABLED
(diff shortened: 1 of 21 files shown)


functionary/core/models/function.py
```diff
@@ -80,9 +91,23 @@ def deactivate(self):
         """Deactivate the function and pause any associated scheduled tasks"""
         self.active = False
         self.save()
+        self.update_scheduled_tasks(enable=False)
 
+    def update_scheduled_tasks(self, enable: bool) -> None:
```
scott-taubman, line 96: The two different behaviors here feels like belong in two separate functions.  But also, I don't think calling this with enable = True would ever be a good idea, so I'd probably just eliminate that path and make this `pause_scheduled_tasks`.

functionary/core/models/function.py
```diff
@@ -99,3 +124,7 @@ def render_name(self) -> str:
     def schema(self) -> dict:
         """Function definition schema"""
         return get_schema(self)
+
+    def is_active(self) -> bool:
+        """Returns true if active and the package isn't disabled"""
+        return self.active and self.package.enabled
```
scott-taubman, line 130: There was a conversation at some point lamenting our inconsistency with the status terminology.  It would be nice if we could pick one and go with it.  I'd probably vote for `active` over `enabled`.  Even if internally the terms end up being different for some reason, the public facing thing should probably be the same, such that this is:

```python
    @property
    def is_active(self) -> bool:
        """Returns true if active and the package isn't disabled"""
        return self.active and self.package.is_active
```

functionary/core/models/package.py
```diff
@@ -7,6 +7,22 @@
 
 from core.models import Environment
 
+# Pending status represents the package is being built
+# Complete status represents the package has been successfully built at least once
+# Enabled status represents that the package is can be used by users
+# Disabled status represents that the package cannot be used by users
+PENDING = "PENDING"
```
scott-taubman, line 14: I don't love these being bare constants.  I can see why you moved them out here, but I'd rather they be part of some class, similar to `PARAMETER_TYPES` over in `core/utils/parameter.py`.  In cases where you're importing them elsewhere, it provides much improved context IMO.  It also lets you import a single thing and have access to all of the values.

functionary/core/models/package.py
```diff
@@ -55,6 +62,9 @@ class Package(models.Model):
 
     image_name = models.CharField(max_length=256)
 
+    objects = models.Manager()
```
scott-taubman, line 65: Does this need to be defined?  Seeing it here might make one assume that we're changing it away from its default, but that's not actually the case.

functionary/core/models/package.py
```diff
@@ -73,18 +83,25 @@ def update_image_name(self, image_name: str) -> None:
...
     def disable(self) -> None:
         """Disable package and it's associated functions"""
-        # TODO: Add method for disabling package's functions
-        self._update_status(self.DISABLED)
+        # Prevent the user from manually disabling a package that hasn't
+        # yet successfully finished being built.
+        if self.status == ENABLED:
+            self._update_status(DISABLED)
+
+            for function in self.active_functions.all():
```
scott-taubman, line 103: The way I read this, you've already set the status of the package to DISABLED, so `self.active_functions.all()` will always be an empty set, won't it?  I think you want to do this whole thing in a transaction, update the scheduled tasks first, then set the package status.

functionary/core/models/package.py
```diff
@@ -106,3 +123,8 @@ def full_image_name(self) -> str:
     def active_functions(self) -> QuerySet:
         """Returns a QuerySet of all active functions in the package"""
         return self.functions.filter(active=True)
+
+    @property
+    def enabled(self) -> bool:
```
scott-taubman, line 128: Commented elsewhere, but it feels like for the sake of consistency this should be the same as whatever we named the function version of this, which is currently `is_active`.

functionary/core/models/workflow.py
```diff
@@ -96,3 +96,7 @@ def deactivate(self):
         # TODO: Once scheduled_tasks support workflows come back and do something like:
         # for scheduled_task in self.scheduled_tasks.all():
         #     scheduled_task.pause()
+
+    def is_active(self) -> bool:
```
scott-taubman, line 100: I sort of implied this in another comment, but I would just make this a property method.

scott-taubman (changes requested):

scott-taubman (approved):

PR #313, opened 2023-03-30
#302 - Refactor summary and description

### LINKED ISSUE #302: Refactor form pages
Always show summary, but not with a section header. Just `<p>` text under the page header. Show description if it exists, ~maybe with~ without section header. Linebreak description with `| linebreaks`

Area around content should be shaded. remove all section headers

obr42:
Closes #302 

Always show the summary if it exists, in a more condensed way. Next comes description with its section header, if it exists. Previously would show one or the other but not both. Add line-break support to description.


scott-taubman (changes requested):
I think the end result here just makes me feel more strongly that we should drop the section headers.  Also, we talked about other changes to this page and I think trying to do the summary and description changes in isolation is probably not the best approach.  The goal is to clean up the tasking page overall and I think some of the other changes that are needed will probably influence what has to happen with summary and description as well.  Let's take a step back and work on a more comprehensive set of changes to these pages.

PR #321, opened 2023-04-05
#308 - boolean sliders

### LINKED ISSUE #308: Have sliders not check boxes

obr42:
Closes #308 

Use slider indicators in favor of check-boxes. Applied everywhere except function Create Task parameters as that is using default templates which were not configurable.


kvig (commented):
LGTM. Updating the boolean form parameters could be a more involved task than these quick changes.

PR #329, opened 2023-04-06
#327 - tasking type icons in table

### LINKED ISSUE #327: Condense tasking table
Replace type column with corresponding icon in name column

obr42:
Closes #327 

Tasking type in Tasking table is now shown as an icon in the Name column instead of its own column.

### DIFF
--- a/functionary/ui/tables/task.py
+++ b/functionary/ui/tables/task.py
@@ -78,7 +78,7 @@ class TaskListTable(tables.Table):
     def render_name(self, value, record):
         icon = "fa-cube" if record.tasked_type.name == "function" else "fa-diagram-next"
         return format_html(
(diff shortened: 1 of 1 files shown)


functionary/ui/tables/task.py
```diff
@@ -75,7 +75,13 @@ class TaskListTable(tables.Table):
         format=DATETIME_FORMAT, verbose_name="Created", orderable=True
     )
 
+    def render_name(self, value, record):
+        icon = "fa-cube" if record.tasked_type.name == "function" else "fa-diagram-next"
+        return format_html(
+            "<i class='fa {} fa-fw me-2'></i>{}", icon, record.tasked_object.name
```
kvig, line 81: A title would be good here, make sure to initial cap the value.

kvig (changes requested):

kvig (approved):

PR #330, opened 2023-04-06
#322 - Login error navbar fix

### LINKED ISSUE #322: auth error page nav bug
The `templates/socialaccount/authentication_error.html` is loaded when trying to do external login from the main login page. However user is not logged into Functionary so the nav doesn't load correctly. This needs to look better.

obr42:
Closes #322 

When attempting to login with a 3rd party account on the main login page and an error is thrown, the resulting error page had a bad looking nav bar (thin blue line, no contents). It now displays just the Logo and horizontal rule below the logo - nav links are still hidden as user was not authenticated.

### DIFF
--- a/functionary/ui/templates/base.html
+++ b/functionary/ui/templates/base.html
@@ -48,16 +48,16 @@
             <div class="row flex-nowrap">
                 <div class="col-auto px-lg-3 px-2 bg-primary shadow">
                     <div class="d-flex flex-column vh-100 sticky-top">
-                        {% if user.is_authenticated %}
-                            <nav class="navbar navbar-dark nav-pills flex-column mb-auto align-items-start">
-                                <a href="{% url 'ui:home' %}" class="navbar-brand">
-                                    <h1 class="mb-0">
-                                        <span class="display-2">F</span>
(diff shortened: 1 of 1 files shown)


functionary/ui/templates/base.html
```diff
@@ -48,16 +48,16 @@
...
-                                <div class="btn-group d-grid d-md-flex w-100">{% include "forms/environment_selector.html" %}</div>
-                                <hr class="w-100 border-white" />
+                        <nav class="navbar navbar-dark nav-pills flex-column mb-auto align-items-start">
+                            <a href="{% url 'ui:home' %}" class="navbar-brand">
+                                <h1 class="mb-0">
+                                    <span class="display-2">F</span>
+                                    <span class="functionary-title fs-4 d-none d-md-inline">unctionary</span>
+                                </h1>
+                            </a>
+                            <div class="btn-group d-grid d-md-flex w-100">{% include "forms/environment_selector.html" %}</div>
```
kvig, line 58: Since the user isn't authenticated, the environment selector isn't needed here. Either move this into the check below  or wrap the template itself in `environment_selector.html` with the `user.is_authenticated` check also.

kvig (changes requested):

kvig (approved):

PR #332, opened 2023-04-07
277 - Handle exceptions on task create.

### LINKED ISSUE #277: Task Creation Error Handling
If an exception is raised during the creation of a Task, the Task is forever stuck "In Progress", rather than being marked as "ERROR" and logging the error.  We should ensure than all possible exceptions are caught and the Task status and log are updated accordingly.

kvig:
This PR adds some exception handling during the task creation process and the `publish_task` phase. Tasks are marked as `Task.ERROR` and a `TaskLog` is generated or appended to. Appending logs should only happen if `publish_task` fails and celery retries the call.

Closes #277 

To test manually, see the description in #277 to create an invalid workflow and try execute it. Verify that no tasks are left in the `IN PROGRESS` state.


functionary/core/utils/tasking.py
```diff
@@ -250,6 +263,22 @@ def start_task(task: Task) -> None:
...
         _start_workflow_task(task)
     else:
-        raise InvalidContentObject(
-            f"Handling for content type {tasked_type_class} is undefined"
+        message = f"Handling for content type {tasked_type_class} is undefined"
+        task_errored(task, message, None)
+        raise InvalidContentObject(message)
+
+
+def task_errored(task, message, error=None):
```
scott-taubman, line 271: name this something more action-y, inline with `start_task` and `publish_task`.

functionary/ui/views/function.py
```diff
@@ -114,14 +112,37 @@ def execute(request: HttpRequest) -> HttpResponse:
...
             status_code = 503
             form.add_error(
                 None,
                 (
-                    "Unable to upload file. Please try again. "
+                    "Unable to upload file, please try again. "
                     "If the problem persists, contact your system administrator."
                 ),
             )
+        except Exception:
```
scott-taubman, line 124: Let's just let this exception bubble up and not handle it.  If we get here it's because we screwed something up.

functionary/core/utils/tasking.py
kvig: done

functionary/ui/views/function.py
```diff
@@ -114,14 +112,37 @@ def execute(request: HttpRequest) -> HttpResponse:
                 ),
             )
         except S3Error:
+            task = None
             status_code = 503
             form.add_error(
                 None,
                 (
-                    "Unable to upload file. Please try again. "
+                    "Unable to upload file, please try again. "
```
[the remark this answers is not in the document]
kvig: Changed it to a semicolon because I didn't like having 3 sentences.

PR #333, opened 2023-04-11
#324 - table sorting work

### LINKED ISSUE #324: Sorting bugs
- [x] Bug: sorting choice is lost when a filter gets applied
- [x] Bug: sort order is wrong on link fields (possibly sorting by uuid rather than name)
- [x] Sort indicators on every column, different indicator on column that is sorted

obr42:
Closes #324 

Fixes bug so applying filter maintains current sort, if any. Fix bug where sort order was incorrect on Scheduled Task functions.


functionary/ui/templates/partials/table_template.html
```diff
@@ -0,0 +1,31 @@
...
+                            <a href="{% querystring table.prefixed_order_by_field=column.order_by_alias.next %}"
+                               title="Click to toggle sort"
+                               class="text-nowrap">
+                                {{ column.header }}
+                                {% if 'asc' in column.attrs.th.as_html %}
+                                    <i class="fa-solid fa-arrow-up-a-z" title="Sorted in ascending order"></i>
+                                {% elif 'desc' in column.attrs.th.as_html %}
+                                    <i class="fa-solid fa-arrow-down-z-a" title="Sorted in descending order"></i>
+                                {% else %}
+                                    <i class="text-muted fa-solid fa-sort"></i>
```
scott-taubman, line 20: I'm not a fan of having an icon on the non-sorted fields.  I think the fact that the header is clickable (and colored as a link) is enough of an indicator that the field can be sorted.  Or that it can be interacted with at the least, and then the user would learn very quickly what clicking it does.

functionary/ui/templates/partials/table_template.html
```diff
@@ -0,0 +1,31 @@
...
+            <tr>
+                {% for column in table.columns %}
+                    <th {{ column.attrs.th.as_html }} scope="col">
+                        {% if column.orderable %}
+                            <a href="{% querystring table.prefixed_order_by_field=column.order_by_alias.next %}"
+                               title="Click to toggle sort"
+                               class="text-nowrap">
+                                {{ column.header }}
+                                {% if 'asc' in column.attrs.th.as_html %}
+                                    <i class="fa-solid fa-arrow-up-a-z" title="Sorted in ascending order"></i>
```
scott-taubman, line 16: This should be `fa-arrow-down-a-z`.  If both the arrow and the letters change direction, then the meaning of the two icons feels the same.  I figured I'd check what someone else does, with my first instinct being Excel, and sure enough:

image

Having said that, I don't personally feel like this icon is an improvement over the old one.  Seeing the AZ or ZA on columns that are not alphabetical listings (which I guess is just dates) feels odd to me, but also doesn't feel like it conveys any more that a simple up or down indicator.  If you really hate the chevron, I think `sort-up` / `sort-down` could look fine (though those are difficult to align, which is why I originally chose the chevron) as would plain old `arrow-up` / `arrow-down`

scott-taubman (changes requested):

functionary/ui/tables/schedule_task.py
```diff
@@ -35,6 +35,7 @@ class ScheduledTaskTable(tables.Table):
         ),
         verbose_name="Scheduled Task",
     )
+    function = tables.Column(accessor="function__name")
```
scott-taubman, line 38: With the accessor change you need to manually set the name, otherwise it will be "Name" rather than "Function".  

```suggestion
    function = tables.Column(accessor="function__name", verbose_name="Function")
```

PR #334, opened 2023-04-11
#312 - human readable crontab

### LINKED ISSUE #312: Format cron date in table
Schedules table

obr42:
Closes #312 

Human readable crontab format on table and in schedule task details page in a Popover. Had to update celery to latest version, and add Popover support


scott-taubman (changes requested):
Displaying the human readable text in place of the crontab syntax does not feel like an ideal solution to me for a couple of reasons:

* It sort of obfuscates the actual value that is set
* I'm guessing a good portion of the intended user base is sufficiently familiar with crontab syntax that they would find that quicker and easier to parse at glance.
* The human readable form gets very wordy and results in the "Last Run" data getting broken up into multiple lines, which looks silly:

image

What would work better IMO is to put the human readable form in hover over text, either on the crontab value itself or via an info icon.  That way the actual set value is what gets displayed, but a friendlier version is available.  The bootstrap [popover](https://getbootstrap.com/docs/5.3/components/popovers/) could be a solution for this.

scott-taubman (approved):

PR #350, opened 2023-04-17
#337 - change default page to task list

### LINKED ISSUE #337: Make `/ui` go to tasking page

obr42:
Closes #337 

Change the "home" page to be the task list page. Did not change `ui_.py` field `LOGIN_REDIRECT_URL = "ui:home"` or the logo's link `{% url 'ui:home' %}` so that it could easily be changed to a distinct home page in the future. Simply made the "home" page re-direct to the task list page.


scott-taubman (changes requested):
The home view was doing the work of choosing the initial environment when you login.  With that gone, there is no environment set in the session after login and you end up with something of a broken site if you're an admin and a 403 error if you're a regular user.

Two possible approaches here:

* Keep the home view, have it do what it was doing, but then redirect at the end instead of rendering the `home.html` template.  That's slightly hacky, but also leaves the plumbing in for when we get around to actually creating a proper landing page (which I believe we should do, just not now).
* Remove the home view as you have here, but update the login process so that the initial environment setting happens as part of a successful login.  Or possibly update the session handling so that we always set the session if it's not set. That seems like the more robust solution long-term, but I couldn't tell you off hand how to do that.

scott-taubman (approved):

PR #352, opened 2023-04-19
Streamline crontab display in schedules list

scott-taubman:
This PR streamlines the crontab display on the list page by removing the " (m/h/dM/MY/d) UTC" format suffix.  This was done for a few reasons:

* It was noisy and only questionably helpful
* The popover text that gives the human readable form feels like it makes that format helper text redundant anyway
* It caused weird wrapping on even moderately sized windows

Additionally, the order of the crontab input fields on the schedule create / edit page has been fixed to match the actual order of crontab values (both the standard syntax and the way the values are displayed on every other page).


kvig (approved):
LGTM

